PR #358, opened 2024-08-23
Add Automated Script to Reset Subscription Usage Count Every 30 Days

### LINKED ISSUE #355: Script to refresh usage count
### What is this issue related to?

Enhancement

### Description

This task involves creating a script which runs everyday at midnight and checks the subscriptions collection to refresh the usage count.

### Expected Behavior

1) For every subscription where the usage Count has hit the usage limit, or if the monthly time duration for subscription has elapsed, the usage count must be reset to zero.
2) This script must run at midnight everyday to scan the database and refresh the usageCount.

### Current Behavior

There is no such script as of now.

### Additional Information (optional)

_No response_

### Confirmation Checklist

- [X] I have searched for similar issues before creating this one.
- [X] I have provided all the necessary information to understand and reproduce the issue.

DeltaDynamo:
### Summary

This pull request introduces a new script that is scheduled to run automatically at midnight to refresh the usage count of subscriptions. 

### Description

The script performs the following actions:
- Connects to the MongoDB database and retrieves all subscriptions.
- Calculates the number of days since each subscription was created.
- Resets the usage count to zero for subscriptions after every 30 consecutive days starting from the time of creation and only if the current usage count is greater than zero.
- Updates the `updatedAt` field to reflect the reset action.

This PR resolves [Issue #355](https://github.com/TeamShiksha/logoexecutive/issues/355)

### Screenshots or Recordings (Optional)

Screenshot 2024-08-23 234213

## Checklist

- [X] I have tested the changes locally and they work as expected.
- [X] This pull request follows the project's coding guidelines.

### DIFF
--- a/server/scripts/refreshUsageCount.js
+++ b/server/scripts/refreshUsageCount.js
@@ -1,0 +1,34 @@
+const { Subscriptions } = require("../models");
+const mongoose = require("mongoose");
+require("dotenv").config();
+
+async function refreshUsageCount() {
+  try {
+    await mongoose.connect(process.env.MONGO_URL);
+
+    const allSubscriptions = await Subscriptions.find();
+    
+    const refreshSubscriptionData = allSubscriptions.map(async (subscription) => {
+      const currentDate = Date.now();
+      const daysSinceLastUpdated = (currentDate - subscription.updatedAt) / (1000 * 3600 * 24);
+
+      if (subscription.usageCount >= subscription.usageLimit || daysSinceLastUpdated >= 31) {
+        subscription.usageCount = 0;
+        subscription.updatedAt = Date.now();
+        await subscription.save();
+      }
+    });
+
+    await Promise.all(refreshSubscriptionData);
+  } catch(err) {
+    console.log(err);
+  } finally {
+    await mongoose.connection.close();
+  }
+}
+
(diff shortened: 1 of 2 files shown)


server/scripts/refreshUsageCount.js
aps08, line 1: @amankumarsingh77 can you confirm if we use promise all. The number of API call per second will hit the limit for the mongoDB cluster. Consider that the number of users > 500. @DeltaDynamo you might need to change the approach if it hits the limit.

aps08: Also remove the comments

DeltaDynamo: Hi Anoop, you are right. I will look into it and update the code accordingly.

DeltaDynamo:
Hi @aps08 Following your last comment. I have made the following changes.

- I have removed individual save for records, instead of it, now using bulkWrite() to update multiple records at once. This will reduce the unnecessary API calls that was earlier being made. As a result, promise.all() was also not required so removed it as well.
- I have updated the find query to fetch only those records where usageCount > 0 because anyway we are going to update only those records where usageCount is greater than 0.

Kindly review the changes made. Let me know if you find any issues with the current approach.

aps08 (approved):

Issue #368, opened 2024-09-20
 Implement Proper Template for "Forget Password"


sujal111:
### What is this issue related to?

Enhancement

 Implement Proper Template for "Forget Password"

**Description:**

When users forget their credentials, they should have the ability to reset their password via the "Forget Password" option. Currently, while this option exists, the reset email does not have an appropriate template, and it lacks a proper greeting or salutation for the customers.

### Expected Behavior

- When a user clicks on the "Forget Password" option, they will receive an email with a well-structured template that includes:
  - A proper salutation ,
  - Clear instructions to reset their password
  - Professional and user-friendly messaging
  - A link to reset the password

### Current Behavior

- The "Forget Password" functionality is available, but the email template does not include a user-friendly greeting or any professional salutation, creating a less-than-ideal user experience.

### Additional Information (optional)

**Steps to Reproduce:**
1. Go to the login page.
2. Click on the "Forget Password" option.
3. Observe the email received for resetting the password.

### Confirmation Checklist

- [ ] I have searched for similar issues before creating this one.
- [X] I have provided all the necessary information to understand and reproduce the issue.

aps08:
@sujal111, why is this not closed ?

PR #372, opened 2024-09-26
Feature: HTML email template

asharma991:
### Summary

#368

This PR implements HTML templates for sending out emails to the users of LogoExecutive. Currently, we're just sending plain emails with very little content. So, this change will improve the communication and brand image.

### Approach

Multiple approaches were taken to create the templates. I made plain HTML templates using some CSS. However, after testing them on various clients, it was observed that the behavior of plain HTML structure is quite different for different clients. So to make it consistent, table structure is used and instead of using classes inline styles are used. (A good guide for creating HTML templates: https://www.freecodecamp.org/news/how-to-create-a-responsive-html-email-template/)

### How to Test the Changes

Try sending out an email using operator credentials or click on forgot password. You'll receive the email in the new templates

### Screenshots or Recordings

These are the screenshots taken for multiple clients

**Before**:
image
image
image

**After**:
image
image
image

## Checklist

- [x] I have added/updated tests that cover the changes.
- [x] I have updated the documentation to reflect the changes.

### DIFF
--- a/server/__tests__/unit/utils/sendEmail.js
+++ b/server/__tests__/unit/utils/sendEmail.js
@@ -36,6 +36,7 @@ describe("sendEmail", () => {
       to: "test@example.com",
       subject: "Test Subject",
       text: "Test Body",
+      html: "Test Body",
     });
     expect(result).toEqual({ success: true });
   });
@@ -71,6 +72,7 @@ describe("sendEmail", () => {
       to: "test@example.com",
       subject: "Test Subject",
       text: "Test Body",
+      html: "Test Body",
     });
     expect(result).toEqual({ success: false, error: new Error("Test Error") });
   });
--- a/server/controllers/auth/forgot-password.js
+++ b/server/controllers/auth/forgot-password.js
@@ -1,85 +1,9 @@
 const Joi = require("joi");
 const { STATUS_CODES } = require("http");
+const fs = require("fs");
+const handlebars = require("handlebars");
 const { fetchUserByEmail, createForgotToken } = require("../../services");
 const { sendEmail } = require("../../utils/sendEmail");
-
-const getHTMLBody = (url) =>{
-  return `<!DOCTYPE html>
-    <html lang="en">
-      <head>
-        <meta charset="UTF-8">
-        <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <title>Forgot Password</title>
-        <link rel="preconnect" href="https://fonts.googleapis.com" />
-        <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
-        <link
-          href="https://fonts.googleapis.com/css2?family=Inter:wght@300;400;500;600;700;800;900&family=Lato:wght@700&family=Nunito&family=Poppins&display=swap"
-          rel="stylesheet" />
-      </head>
-
-      <body style="margin: 0; padding: 0; font-family: Inter, Arial, sans-serif; box-sizing: border-box;">
-        <table width="100%" cellpadding="0" cellspacing="0"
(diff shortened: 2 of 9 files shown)


aps08:
Can we get the latest screenshots ? There were changes suggested in the meeting. @asharma991

aps08 (changes requested):
HTML templates should be in .html files not directly inside the code. You should import these html files and email the user.

asharma991:
> HTML templates should be in .html files not directly inside the code. You should import these html files and email the user.

But these files have URLs and dynamic content that comes from those respective utilities. Should I create separate functions for these?

aps08:
@asharma991 updates ?

server/__tests__/unit/utils/sendEmail.js
```diff
@@ -36,6 +36,7 @@ describe("sendEmail", () => {
       to: "test@example.com",
       subject: "Test Subject",
       text: "Test Body",
+      html: "Test Body",
```
Ayushsanjdev, line 39: Is this test data required?

server/__tests__/unit/utils/sendEmail.js
```diff
@@ -71,6 +72,7 @@ describe("sendEmail", () => {
...
```
Ayushsanjdev, line 75: +1

server/controllers/auth/forgot-password.js
```diff
@@ -46,8 +43,18 @@ async function forgotPasswordController(req, res, next) {
...
-    const mail = mailText(userToken.tokenURL());
-    const nodeMailerRes = await sendEmail(user.email, mail.subject, mail.body);
+    const htmlFile = fs.readFileSync(
+      __dirname + "/../../templates/ForgotPasswordOrVerify.html",
+      "utf-8"
+    ).toString();
+    const template = handlebars.compile(htmlFile);
+    const replacements = {
+      url: userToken.tokenURL(),
+      message1: "It seems like you forgot your password?",
```
Ayushsanjdev: Let make text dynamic or add in contants.

server/package.json
```diff
@@ -1,39 +1,40 @@
...
+  },
+  "devDependencies": {
+    "@shelf/jest-mongodb": "^4.3.2",
+    "@types/jest": "^29.5.6",
+    "eslint": "^8.46.0",
+    "jest": "^29.6.4",
+    "nodemon": "^3.0.1",
+    "supertest": "^6.3.3"
+  }
+}
```
Ayushsanjdev, line 40: What is the change in package.json

Ayushsanjdev (changes requested):
If this PR is not urgent, you can do the changes.
Rest looks good to me

server/__tests__/unit/utils/sendEmail.js
asharma991: Yup it is, we've added the templates to send the HTML response hence we need this

server/package.json
asharma991: A new dependency handlebars is added for replacements in the HTML templates

server/controllers/auth/forgot-password.js
asharma991: This is only used in this file. Moving it to constants will make sense only when this is used at several places

server/controllers/auth/signup.js
```diff
@@ -106,10 +108,21 @@ async function signupController(req, res, next) {
...
       );
 
+    const htmlFile = fs.readFileSync(
+      __dirname + "/../../templates/ForgotPasswordOrVerify.html",
+      "utf-8"
+    ).toString();
+    const template = handlebars.compile(htmlFile);
+    const replacements = {
+      url: verificationToken.tokenURL(),
+      message1: "You need to verify your email",
```
aps08: Can we change `message1` and `message2` to better variable names

aps08 (changes requested):
nit

aps08 (approved):

Issue #371, opened 2024-09-24
Add log report for SignIn/Up process and store to mongodb


iamchiki:
### What is this issue related to?

Enhancement

### Description

We need to add enhancement for log report creation, whenever User SignIn/SignUp a log reoport is generated and this log report should be stored in mongodb.

### Expected Behavior

A log reoprt should be generated for API route: `api/auth/signin` and `/api/auth/signup`

### Current Behavior

Currently there is no log report generated for API route: `api/auth/signin` and `/api/auth/signup`

### Additional Information (optional)

No response

### Confirmation Checklist

- [X] I have searched for similar issues before creating this one.
- [X] I have provided all the necessary information to understand and reproduce the issue.

PR #379, opened 2024-10-14
Add server logs code in codebase and and log report DB.

iamchiki:
### Summary
This PR add server logs mechanism in codebase, it address issue #371 

### Description

-  Added `Log` model to format log schema in DB.
-  Added `Logs` in `server/services`, it save log report in MongoDB 

### How to Test the Changes

 Add function `await log("log message", http reuest, "info or error");`  importing from `Logs` services in codebase wherever we want to generate log. 

-  1st parameter - it is log message which need to be store in DB.
-  2nd parameter -  it is http req which will give info about ip, host, http method and http url
-  3rd parameter - it is about type of log, If it is good log then parameter is "info" and if it is bad log or error then parameter
    would be "error"

### Screenshots or Recordings

info

errolog

it will create extra `logs` collection in MongoDB and store report as following screen.

mongo

## Checklist

- [x] I have updated the documentation to reflect the changes.

### DIFF
--- a/server/models/Log.js
+++ b/server/models/Log.js
@@ -1,0 +1,29 @@
+const mongoose = require("mongoose");
+
+const LogSchema = new mongoose.Schema({
+  log_message: {
+    type: String,
+    required: true,
+  },
+  user: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: "User",
+    default: null,
+  },
+  request_ip: {
+    type: String,
+  },
+  host: {
+    type: String,
+  },
+  status: {
+    type: String,
+  },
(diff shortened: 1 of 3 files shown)


asharma991 (changes requested):
LGTM, but please write tests for the Logs.js

aps08:
Recently, someone from the other project told me a better approach for log. This might need to be changed. Let's discuss it in our next call.

aps08:
Closing with as we will be using `newrelic`. Storing log in database is not very useful and `newrelic` is giving a generous plan

PR #405, opened 2024-11-15
Business controller repository pattern

Rohan-Hazari:
### Summary

Implemented the repository pattern, abstracted data access logic into repository classes now services and controllers interact with repository interfaces, not directly with the database.

### Approach
Created methods inside Image, Key & Subscription repositories to interact with the database.
These methods were then called in service files and any logical operation with the data was done here and returned
Now finally these methods in service files were used in business controller in search.js and logo.js.
Added cloudFront in util to fetch images signed url .

#### Note
In logo.js and search.js any methods to  check API usage or update it based on userId is not working yet as there is no reference from Key and Subscription model to the User so these will either give you error or return empty data, a comment has been added above such methods

### Output

Endpoint: `business/logo`
You get a single logo which matches the provided domain name

image

Endpoint: `business/search`
You get a list of logo whose name starts with the provided domain name

image

## Checklist


packages/app/controllers/business/logo.js
aps08, line 1: logo.js and search.js should follow the same functionalities.

aps08 (changes requested):
Make changes as requested.

aps08:
To be completed @DeltaDynamo.

Issue #352, opened 2024-08-16
Sign in as a guest


surajmaity1:
### What is this issue related to?

Feature

### Description

Implement "Sign in as a guest" feature that allow users to access limited features of logoexecutice platform without any user registration. 

This feature will be used for giving a brief showcase how our platform works. This session will expire after a set period of time or when user's inactivity.

Requirements:

Endpoint Creation:
 - Create a new API endpoint **/api/auth/guest-signin** for signing in as a guest.

Guest Session Management:

- Generate a unique guest token upon successful guest sign-in.
- Store the guest session in the database with an expiration time (e.g., 24 hours).
- The session should be invalidated if the user decides to sign up or log out.

### Expected Behavior

Access Control:

- Implement access control to ensure guest users can only access specific parts of the application.
 - Define what resources and endpoints are accessible to guest users.

### Confirmation Checklist

- [X] I have searched for similar issues before creating this one.
- [X] I have provided all the necessary information to understand and reproduce the issue.

surajmaity1:
@aps08 assign this issue to me ( @surajmaity1 ) and @TusharAnekar

aps08:
@MukulGupta005

aps08:
@soma2000-lang if you are not working on this let us know.

soma2000-lang:
@aps08 Sorry have been going through many stuff , will give you an update this friday

PR #519, opened 2025-01-30
Add initial code for guest sign in feature

Hariom01010:
### Summary

This PR aims to provide a initial setup for the POC for the Sign In as Guest feature and work forward according to the reviews aiming to achieve the feature
### Approach

Resolving Issue: #352 

On the backend
- A controller has been created which looks for a guest user using the email of the user (already created in DB) and logins using it and sends a JWT token in the cookie as response.
- The email of the guest has to be set using env variables.
- A route has been created which calls this controller.

On the frontend
- A "Continue as guest" span has been created in the signup and signin form.

NOTE: on the frontend we can also set a sign in button instead of the span text we are using (something similar to the OAuth buttons provided by google/facebook, etc)

### How to Test the Changes
The backend api has still not been connected with the frontend. So, setup mongodb and setup a guest user and set its email in env variables. Now in Postman/Thunder Client make a GET Request to http://localhost:5000/api/auth/guest-signin. In case a user is found a 200 status code is sent with setting the jwt token in cookies.

### Screenshots or Recordings

https://github.com/user-attachments/assets/ced07256-bbcb-43c2-82ad-be14e4841e6f

Screenshot 2025-01-30 191849

## Checklist

### DIFF
--- a/packages/app/.env.example
+++ b/packages/app/.env.example
@@ -19,3 +19,4 @@ MONGO_URL=mongodb://admin:password123@mongodb:27017/logoexecutive?authSource=adm
 NODE_ENV=dev
 NEW_RELIC_APP_NAME=LogoExecutive
 NEW_RELIC_LICENSE_KEY=BAB0E5349BDFBFF832FE0C72597EA6333930691A
+GUEST_EMAIL=test@gmail.com
--- a/packages/app/controllers/auth/guest-signin.js
+++ b/packages/app/controllers/auth/guest-signin.js
@@ -1,0 +1,24 @@
+const UserService = require("../../services/User");
+const dayjs = require("dayjs");
+
+async function guestSignInController(req, res, next) {
+    try{
+        const userService = new UserService();
+        const user = await userService.getUserByEmail(process.env.GUEST_EMAIL)
+   
+        if(!user){
+            return res.sendStatus(400).json({message:"Guest user not found"})
+            
+        }
+        res.cookie('jwt', user.generateJWT(), {
+            expires: dayjs().add(1, 'day').toDate(),
+            sameSite: "none",
+            secure: true
+        })
+        res.sendStatus(200)
+    } catch (err){
+        next(err)
+    }
+}
+
+module.exports = guestSignInController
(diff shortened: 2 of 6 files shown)


packages/app/.env.example
```diff
@@ -18,4 +18,5 @@ KEY=s3_bucket_folder_name_can_customise_as_per_your_likings
 MONGO_URL=mongodb://admin:password123@mongodb:27017/logoexecutive?authSource=admin
 NODE_ENV=dev
 NEW_RELIC_APP_NAME=LogoExecutive
-NEW_RELIC_LICENSE_KEY=BAB0E5349BDFBFF832FE0C72597EA6333930691A
\ No newline at end of file
+NEW_RELIC_LICENSE_KEY=BAB0E5349BDFBFF832FE0C72597EA6333930691A
```
lakshay-saini-au8, line 21: @Hariom01010 is this a fake key ?
NEW_RELIC_LICENSE_KEY

Hariom01010: yes sir

lakshay-saini-au8 (approved):

aps08 (changes requested):
Proof submitted only indicates that user exists in the data. Could you provide more details by fetching information about the guest user. Also resolve Sonarcloud issues. Continue as guest should be after the Sign Up.

UI should be something like this in our theme: 

image

Hariom01010:
# Workflow:
Proof
# Signup Form:
signup
# Signin Form:
signin

DaScientist:
Guest login should not be implemented with a test email, as tracking should be there for guest logins too. Better implementation would be to create a separate User type which would be logged whenever a `new` user would request access. Adding multiple entries in the same table increases noise in the data, which is not preferred. Also, regarding route, it would be better to handle this in the sign in api only, and not introduce a separate route.

Another note, while making any UI, it would be preferred if the engineer judges if the UI doesn't seem good, atleast when the UI is not there. The horizontal line below the continue as guest button is looking ugly. The padding is inconsistent which is not looking good at all. Requesting you to look and change these details too.

As these changes are requiring a complete rewrite, and it's been a long time since the PR was created, I'm closing this PR, and suggesting you to do implement these in another branch and raise another PR.

PR #556, opened 2025-02-17
replaced a tag with Link and updated tests

### LINKED ISSUE #549: Replace Anchor (<a>) Tags with React Router <Link> for Client-Side Navigation in header and footer
### Description

The project currently uses <a> (anchor) tags for navigation in the header and footer. This causes a full-page reload and re-downloads the entire bundle, leading to unnecessary performance costs.

### Expected Behavior

Navigation should use <Link> from react-router instead of <a> tags to ensure smooth client-side routing without a full-page reload.

### Steps to Reproduce

1. Open the application in a browser.
2. Click on any navigation link in the header or footer.
3. Observe that the page refreshes completely instead of using client-side routing.
4. Check the browser’s network tab to confirm that the bundle is reloaded unnecessarily.

### Screenshots or Recordings

_No response_

### Confirmation Checklist

- [x] I have searched for similar issues [issues](https://github.com/TeamShiksha/logoexecutive/issues) and made sure no duplicate issue exists..
- [x] I have provided all the necessary information to understand and reproduce the bug.

tanwarAalok:
## Description
#549 Replaced the <a> tag with <Link> tag in Header and Footer and updated the Header and Footer test accordingly.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [x] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🛠️ CI/CD

https://github.com/user-attachments/assets/5598117c-5672-4827-969e-d93faf00629a

image
image

## Checklist


aps08 (changes requested):
Nav is redirecting to different page even when items are on the home page. It should scroll to the section.

tanwarAalok:
@aps08  now navigation is working fine. I have added id to all the section where navigation was there and scrolling there. I have also added ScrolltoTop component to scroll to top whenever route changed because if I was on certain section in home page apart from top and routed to new route like /doc the /doc was not automatically at relative scrolled positioned as it was on previous route.

I didn't understand how should I proceed with testing section navigation as these are not directly part of Header or Footer component. So I left it for now with TODO comment which is why I am getting 2 issue in quality check.

packages/ui/src/components/footer/Footer.jsx
```diff
@@ -1,24 +1,39 @@
...
 import styles from "./Footer.module.css";
+import { Link, useNavigate } from "react-router-dom";
+import { handleSectionClick } from "../../utils/Helpers.js";
 
 function Footer() {
+  const navigate = useNavigate();
+
   return (
-    <div className={`container ${styles.block}`}>
+    <div role="footer" className={`container ${styles.block}`}>
```
aps08: what is role here ?

packages/ui/src/components/header/Header.jsx
```diff
@@ -23,7 +24,7 @@ const Header = () => {
   };
 
   return (
-    <div className={`container ${styles.block}`}>
+    <div role="header" className={`container ${styles.block}`}>
```
aps08: what is role here ?

packages/ui/src/utils/Constants.js
```diff
@@ -94,7 +94,7 @@ export const HEADER_ITEMS = [
   {
     name: "home",
     title: "Home",
-    url: "/",
+    url: "#home",
```
mayank2424, line 97: What's the reason to change it to `#`?

tanwarAalok: If a user navigates to other sections like Features or Pricing and then clicks the Home button, it should scroll back to the top of the page. Using "/home" won't work in this case because the user is already on the homepage.

packages/ui/src/components/footer/Footer.jsx
tanwarAalok: I am using role so that I can point out to footer in test 

``` const footerElement = screen.getByRole("footer");```

aps08:
pre commit failing

tanwarAalok:
> pre commit failing

fixed

PR #588, opened 2025-03-03
Ubuntu upgrade in pre-commit - Testing

### LINKED ISSUE #583: Upgrade Ubuntu version and verify pre-commit
We recently received the notification given in the image, that ubuntu version we are using might be decommissioned in future so we need to upgrade the version. Upgrading the version comes with the cost that some libraries are not available which is making it difficult to run the local mongo server. We have to update the pre-commit so that these dependencies should work.

Image

tanwarAalok:
## Description
#583 

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🛠️ CI/CD

## Checklist
- [ ] I have performed a self-review of my code  


aps08 (changes requested):
- 12 file changes for what ? It should have only 1 file change I believe.
- no description, no explanation of the task
- no verification of the task or proof

PR #593, opened 2025-03-05
Adds automated testcases for users functionalities.

### LINKED ISSUE #570: Addition of automated testcases for Users functionalities.
### Describe the enhancement

The Enhancement includes addition of automated testcases for users functionalities such as:

- Retrieving users data
- Updating users data
- Deleting users data
- Generating required keys
- Deleting the keys
- Requesting the logos 

### Why should this enhancement be implemented? Add additional information.
These enhancement should be implemented in order to ensure maintainability of the users functionalities. 

### Confirmation

Severin84:
## Description
closes #570 : 
1. This PR includes addition of automated testcases for the users functionalities such as : 

- functionality for getting users data
- Updating users data
- Deleting users data 
- Generating the required keys 
- Deleting the required keys 
- functionality for requesting logo.

2. This PR also includes functionality improvements such as : 

- check for received users data 
- check for received subscriptions data
- Addition of RequestService in index file.
- Removal of unwanted await in password update functionality 
- Removal or __v from receiving user or subscription data 

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [x] 👨‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🛠️ CI/CD

image
image

## Checklist


DaScientist:
Please update and resolve the sonarcloud analysis issue

packages/app/repositories/base.js
aps08, line 1: please explain this change

Severin84: This removes __v : 0 ,  from the user response or subscription response when we will get them from MongoDB.

packages/app/controllers/users.js
```diff
@@ -231,7 +263,7 @@ async function updatePasswordController(req, res, next) {
       });
     }
 
-    const userUpdateSuccessful = await userService.updateUserPassword(
+    const userUpdateSuccessful = userService.updateUserPassword(
```
aps08, line 259: why `await` is remove from here ?

packages/app/controllers/users.js
```diff
@@ -147,9 +147,25 @@ async function generateKeyController(req, res, next) {
...
+        statusCode: 404,
+        error: STATUS_CODES[404],
+        message: Messages.USER_NOT_FOUND,
+      });
+    }
+
     const subscription = await subscriptionService.getSubscription(
       user.subscription_id
     );
+    if (!subscription) {
```
aps08: not required here. subscription will always exists if signup was successfully. If the signup was not successful then user will not be able to signin and hence won't be able to make api call to this controller.

packages/app/controllers/users.js
Severin84: I tried the update password and it results in an error with the current code because,  **await** is needed on line 60 under app/services/users.js , as we are doing an **async** operation there. And with the current code  this **await** has no effect and since we don't have **await** on line 60 in app/services/users.js it results in error , to fix this I have added **await** in line 60 under app/services/users.js and omitted **await** from here.

packages/app/controllers/users.js
Severin84: So should I remove this ? I added this check to cover a case where user might query a subscription_id and that subscription may or may not exist.

aps08: user is related to subscription right ? Don't we have ref here ?

Severin84: Yes we do have ref.

Severin84: removed it.

aps08 (approved):

PR #614, opened 2025-03-22
Guest Sign In Feature Implementation

### LINKED ISSUE #352: Sign in as a guest

Hariom01010:
## Description

Fixes: #352 

This PR aims to implement the Guest Sign In Functionality. 
- A dummy user similar to other users will be added to the Users Model
- Whenever a user logins as guest, we send a token on the frontend in cookies.
- When the user is logged in as guest, the dashboard buttons are disabled so that user can't make any request to the server. 
- Test case is added to ensure the user can successful sign in as guest.

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [X] ✅ Test
- [ ] 🛠️ CI/CD

### Dashboard when user is logged in as guest:
https://github.com/user-attachments/assets/ade9fd90-d302-4371-8511-263810932ac2

### Flow of Guest Sign In:
Screenshot 2025-03-31 181717

### Test Results:
image

## Checklist
- [X] I have performed a self-review of my code  
- [X] I have commented my code, particularly in hard-to-understand areas  
- [X] I have added tests that prove my fix is effective or that my feature works  
- [X] New and existing unit tests pass locally with my changes


packages/app/controllers/auth.js
```diff
@@ -102,45 +105,82 @@ async function signinController(req, res, next) {
...
-      return res.status(404).json({
-        error: STATUS_CODES[404],
-        message: Messages.INCORRECT_EMAIL_PASS,
-        statusCode: 404,
+    const { email, password, isGuest } = value;
+
+    if (isGuest) {
+      const guestDetails = {
+        ip: req.ip,
+        token: req.cookies["x-guest-token"],
```
mayank2424: Please move the key to a constant file.

packages/app/controllers/auth.js
```diff
@@ -102,45 +105,82 @@ async function signinController(req, res, next) {
...
+        return res.status(createGuestUser.status).json({
+          error: STATUS_CODES[createGuestUser.status],
+          message: createGuestUser.message,
+          statusCode: createGuestUser.status,
+        });
+      } else if (
+        createGuestUser.status === 403 ||
+        createGuestUser.status === 404
+      ) {
+        res.clearCookie("x-guest-token", {
```
mayank2424: Move the key name to a constant here as well.

packages/app/controllers/auth.js
```diff
@@ -102,45 +105,82 @@ async function signinController(req, res, next) {
...
+        });
+        return res.status(200).json({ statusCode: 200 });
+      }
+    } else {
+      const user = await userService.getUserByEmail(email);
+      if (!user)
+        return res.status(404).json({
+          error: STATUS_CODES[404],
+          message: Messages.INCORRECT_EMAIL_PASS,
+          statusCode: 404,
```
mayank2424, line 114: Could you please move this to some common function, something like this? It would be great if you could do this for other response codes too.

```
const notFoundResponse = (res: Response, message?: Imessage) => {
    return res.status(404).json({
        error: STATUS_CODES[404],
        message: Messages.INCORRECT_EMAIL_PASS,
        statusCode: 404,
    })
 }
```

packages/app/controllers/auth.js
```diff
@@ -102,45 +105,82 @@ async function signinController(req, res, next) {
...
+      const matchPassword = await user.matchPassword(password);
+      if (!matchPassword) {
+        return res.status(404).json({
+          error: STATUS_CODES[404],
+          message: Messages.INCORRECT_EMAIL_PASS,
+          statusCode: 404,
+        });
+      }
+      const currentDate = new Date();
+      const tomorrow = new Date(currentDate.getTime() + 24 * 60 * 60 * 1000);
```
mayank2424: Rename this to a readable format, something like `oneDayValidityTimestamp`

packages/app/repositories/guestuser.js
```diff
@@ -0,0 +1,31 @@
...
+class GuestUserRepository extends BaseRepository {
+  constructor() {
+    super(GuestUsers);
+  }
+
+  /**
+   * Find a user by Device ID
+   * @param {string} id - The device ID to search for
+   */
+
```
mayank2424: Please fix spacing issue

packages/app/repositories/guestuser.js
```diff
@@ -0,0 +1,31 @@
...
+    const guest = await this.model.findOne({ deviceID: hashedDeviceID });
+    return guest || null;
+  }
+  async deleteUserByDeviceID(hashedDeviceID) {
+    const deletedGuest = await this.model.deleteOne({
+      deviceID: hashedDeviceID,
+    });
+    return deletedGuest || null;
+  }
+}
```
mayank2424: Keep function argument name same as function name

```typescript
  async deleteUserByDeviceID(deviceId) {
    const deletedGuest = await this.model.deleteOne({
      deviceID: hashedDeviceID,
    });
    return deletedGuest || null;
  }
```

packages/app/schemas/auth.js
```diff
@@ -1,18 +1,27 @@
...
-      "any.required": "Email is required",
-      "string.pattern.base": "Invalid email",
+  isGuest: Joi.boolean().default(false),
+  email: Joi.when("isGuest", {
+    is: true,
+    then: Joi.forbidden(),
+    otherwise: Joi.string()
+      .trim()
+      .required()
+      .regex(/^\w+([-+.']\w+)*@\w+([-.]\w+)*\.\w+([-.]\w+)*$/)
```
mayank2424: Please move this regex pattern to a constant

packages/app/services/guestuser.js
```diff
@@ -0,0 +1,59 @@
...
+        const decodedToken = jwt.verify(token, process.env.JWT_SECRET);
+        const guest = await this.guestUserRepository.findUserByDeviceID(
+          decodedToken.id
+        );
+        if (guest) {
+          return {
+            status: 409,
+            message: "Guest User already logged in",
+            token,
+          };
```
mayank2424: service function should only contain the business logic and should not dictate the response structure.

packages/app/services/guestuser.js
```diff
@@ -0,0 +1,59 @@
...
+      { id: guest.deviceID, role: guest.role },
+      process.env.JWT_SECRET,
+      {
+        expiresIn: "2h",
+      }
+    );
+    return { message: "Guest Signed In", token };
+  }
+
+  async deleteUserAccount(id) {
```
mayank2424: Rename this to `deleteAccountById`

packages/ui/src/components/auth/Signin.jsx
```diff
@@ -48,6 +48,9 @@ const SignIn = ({ toggleForm }) => {
     setFocusedField(null);
   };
 
+  const guestSignIn = () => {
+    console.log("Guest Sign in initiated");
```
mayank2424: Remove log?

packages/app/controllers/auth.js
Hariom01010: Should i move the token name to the env file and use that env wherever the token is being used?

packages/app/controllers/auth.js
Hariom01010: This code was already present in the codebase, but i have moved it to the else block. So should i do this in this PR only or is this out of scope for this PR

packages/app/controllers/auth.js
Hariom01010: Same for this, this variable wasn't introduced in this PR. Should i rename this or is this out of scope for this PR

packages/ui/src/components/auth/Signin.jsx
Hariom01010: I have introduced the function where the api call code will be introduced in future. Should i just update it to an empty function ?

packages/app/controllers/auth.js
Hariom01010: Moved it to constants.js

packages/app/controllers/auth.js
Hariom01010: Moved it to constants.js

packages/app/controllers/auth.js
Hariom01010: Renamed the variable since it was a small change

packages/app/repositories/guestuser.js
Hariom01010: Fixed the spacing

packages/app/repositories/guestuser.js
Hariom01010: Updated the argument name

packages/app/schemas/auth.js
Hariom01010: Moved this regex pattern to constants.js

packages/app/services/guestuser.js
Hariom01010: Renamed the function

packages/ui/src/components/auth/Signin.jsx
Hariom01010: Updated the function to an empty function and removed the console.log

packages/ui/src/components/auth/Signin.jsx
```diff
@@ -48,6 +48,8 @@ const SignIn = ({ toggleForm }) => {
     setFocusedField(null);
   };
 
+  const guestSignIn = () => {};
```
DaScientist: Why the empty function?

packages/ui/src/contexts/AuthContext.jsx
```diff
@@ -31,8 +35,14 @@ export const AuthProvider = ({ children }) => {
...
-        [isAuthenticated, setIsAuthenticated, logout]
+        () => ({
+          isAuthenticated,
+          setIsAuthenticated,
+          isGuest,
+          setIsGuest,
+          logout,
+        }),
+        [isAuthenticated, setIsAuthenticated, isGuest, setIsGuest, logout]
       )}
```
DaScientist: Don't use useMemo directly, extract this into a variable

DaScientist (changes requested):

DaScientist:
Disable all the other inputs too when guest login, i.e. email and password change and all such fields

DaScientist:
Does continue as guest redirect to dashboard? it should navigate to dashboard

Hariom01010:
> Does continue as guest redirect to dashboard? it should navigate to dashboard

Currently the dashboard is under a protected route, it checks if the user is authenticated or is a guest. So should i add the navigation to dashboard in the  `const guestSignIn = () => {};` function and set a dummy value for the cookie used to pass the authentication to check if the user is a guest or not or just implement Navigation logic and leave the authentication check for future?

packages/ui/src/components/auth/Signin.jsx
Hariom01010: Will implement the Navigation logic in this as suggested by you

packages/app/services/guestuser.js
Hariom01010: I have moved the response from service file to controller file and the srevice function now just returns the values.

packages/ui/src/components/auth/Signin.jsx
Hariom01010: Implemented the navigation logic in this function

packages/ui/src/contexts/AuthContext.jsx
Hariom01010: Moved the useMemo to a variable

aps08 (changes requested):
- Do we need to add a new model itself for such a small feature ?
- Can't we add a user in the users model for user_type as GUEST ?
- You are using document.cookie to access cookies but it won't work. We are using http only cookies which doesn't allow read using JavaScript.
- What is the use of making the input field read-only, if you are also disabling the submit button ? Even if we allow guest users to edit the input field they won't be able to submit.

Hariom01010:
> * Do we need to add a new model itself for such a small feature ?
> * Can't we add a user in the users model for user_type as GUEST ?
> * You are using document.cookie to access cookies but it won't work. We are using http only cookies which doesn't allow read using JavaScript.
> * What is the use of making the input field read-only, if you are also disabling the submit button ? Even if we allow guest users to edit the input field they won't be able to submit.

Hariom01010:
> * Do we need to add a new model itself for such a small feature ?
> * Can't we add a user in the users model for user_type as GUEST ?
> * You are using document.cookie to access cookies but it won't work. We are using http only cookies which doesn't allow read using JavaScript.
> * What is the use of making the input field read-only, if you are also disabling the submit button ? Even if we allow guest users to edit the input field they won't be able to submit.

- I have gone with a different model, because of 2 major points, first the schema for the guest user and a normal user is very different and second these logins are temporary we need to delete these users after 2 hours for which we are using ttl on the schema which helps us control the deletion of the guest user documents after 2 hours
- We can but then we need to figure out how we will delete specifically the documents with user_type as GUEST after 2 hrs of creation of that document.
- For document.cookie since we can't access cookies. We are authorising the token in the controller. We will use that check to send a authorised:true response which can use to give restricted access.
- You are right, since the buttons are already disabled we can allow editing the input to help the guest user get more feel of how the dashboard will work.

packages/app/.env.example
```diff
@@ -7,6 +7,7 @@ NEW_RELIC_APP_NAME=openlogo
 NEW_RELIC_LICENSE_KEY=BAB0E5349BDFBFF832FE0C72597EA6333930691A
 EMAIL_SERVICE_URL=https://tsp-emailservice.vercel.app/email
 EMAIL_SERVICE_AUTH_TOKEN=gsdAlKTsj6J2x6BVkw3jCPPZij
+GUEST_USER=guestuser@gmail.com
```
aps08: not requried

packages/app/controllers/auth.js
```diff
@@ -102,45 +103,59 @@ async function signinController(req, res, next) {
...
 
-    if (!user.is_verified)
-      return res.status(403).json({
-        error: STATUS_CODES[403],
-        message: Messages.EMAIL_NOT_VERIFIED,
-        statusCode: 403,
-      });
+    if (isGuest) {
+      const guestUser = await userService.getUserByEmail(
+        process.env.GUEST_EMAIL
```
aps08: emails not required. There would be only 1 user with USER type as guest

packages/app/.env.example
Hariom01010: Removed this, since a new function has been introduced in the User Repository to get guest user.

packages/app/controllers/auth.js
Hariom01010: Remove this and introduced a new function in the user repository to get the guest user. Replaced this with that function call.

Hariom01010: Also updated the test coverage after the changes

packages/app/__tests__/controllers/auth/signin.js
```diff
@@ -10,10 +10,12 @@ const app = require("../../../server");
 describe("SIGNIN API", () => {
   beforeAll(() => {
     process.env.JWT_SECRET = "jwtsecret";
+    process.env.SALT_ROUNDS = 10;
```
aps08: remove

aps08:
@Hariom01010 now you can proceed with this.

Hariom01010:
> @Hariom01010 now you can proceed with this.

Yes will be picking it up again

aps08 (commented):
need to see live demo

aps08 (changes requested):
requested changes as discussed.

Hariom01010:
> requested changes as discussed.

I have transferred the boolean value to parent container and passed it on as prop to children and have removed this option from Signup form as discussed

aps08 (approved):

Issue #615, opened 2025-03-24
Integrating APIs


iamalhera:
### Describe the feature

Integrating following backend APIs with frontend screens-
1. gettingUserData
2. signup
3. signin
4. signout

task is carried out by @shafat730 @farhan294sha @Severin84 

### Confirmation

- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand this documentation issue.

Issue #562, opened 2025-02-17
Add functionality to signup Admins.


Severin84:
### Describe the enhancement

Admin should signup with "Admin" as the role and other users should signup as "CUSTOMER" as the role.

![Image](https://github.com/user-attachments/assets/2aa7be62-7e1c-449c-8f2f-24ca4ecc27a0)

### Why should this enhancement be implemented? Add additional informaiton.

Currently all the users are signing up as "CUSTOMER" as the role so currently there is no segregation between users and the Admins. And in future we might need to give Admins some extra functionalities so this enhancement will help us in identifying the Admins.  

### Confirmation

- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand this documentation issue.

PR #625, opened 2025-04-08
Signup integration

Severin84:
## Description
 Contributes to #615 : Integrates Signup api. Covers the functionalities and test cases both. 
 The signup route is implemented , the user can signup and after successful signup user will be toggled to signin component.
 This Pr also includes changes in .env.example file:
-  .env.example(ui) : BASE_URL is changed to VITE_BASE_URL , reason being url was not accessible.  
-  .env.example(app): Included ADMINSEMAILS as it was created issue and was not includes by me in #562 .

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [x] 👨‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🛠️ CI/CD

Screenshot 2025-04-09 050019

## Checklist

### DIFF
--- a/packages/ui/__tests__/components/auth/Signup.test.jsx
+++ b/packages/ui/__tests__/components/auth/Signup.test.jsx
@@ -5,6 +5,8 @@ import {
   SIGNUP,
   PASSWORD_VALIDATION_MESSAGES,
 } from "../../../src/utils/Constants";
+import { renderHook } from "@testing-library/react";
+import { useApi } from "../../../src/hooks/useApi";
 
 describe("SignUpForm UI and Functionality Tests", () => {
   it("renders all form elements correctly", () => {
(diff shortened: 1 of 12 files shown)


packages/ui/src/components/auth/Signup.jsx
```diff
@@ -47,6 +53,13 @@ function SignUp({ toggleForm }) {
     setFormErrors({});
     setIsSubmit(false);
     setFocusedField(null);
+
+    const success = await makeRequest();
+    if (success) {
```
mridxl, line 58: We should show the user some feedback on the UI if signup request fails, if the email is already registered, etc

mridxl (changes requested):

packages/ui/src/api/api_instance.js
```diff
@@ -1,8 +1,7 @@
 import axios from "axios";
 
 export const instance = axios.create({
-  // baseURL: import.meta.env.BASE_URL,
-  baseURL:"http://localhost:5000",
+  baseURL: `${import.meta.env.VITE_BASE_URL}/api/`,
```
mridxl: IMO we should remove the trailing `/` from the baseURL and then start the path from `/` when making api calls in components.
Axios omits the trailing `/` in the baseURL in the [docs](https://axios-http.com/docs/config_defaults)

packages/ui/src/components/auth/Signup.jsx
Severin84: Sure , I will push this update once done.

packages/ui/__tests__/components/auth/Signup.test.jsx
```diff
@@ -5,6 +5,8 @@ import {
   SIGNUP,
   PASSWORD_VALIDATION_MESSAGES,
 } from "../../../src/utils/Constants";
+import { renderHook } from "@testing-library/react";
```
mridxl: You have two import statements, importing from the same library. Please combine them.

Severin84: Done.

PR #738, opened 2025-06-23
Integrated admin dashboard

### LINKED ISSUE #692: Integrate Admin Catalog API
### Describe the feature

Admin Catalog API are available and needs to be integrated with the frontend. Also Add Image Button will be fixed as part of this issue.

As of now we are showing dummy data for admin catalog and hence we need to remove that and integrate the API

Image
The above image contains dummy data.

### Confirmation

tanwarAalok:
## Description

Issue #692
 
Added the Admin Catalog API into the Admin Dashboard and connected it through the dashboard route. Also, cleaned the Protected Route component.

- interagrated catalog api in Admin dashboard
- added max-height in `add image` button in admin dashboard
- removed search icon from dashboard which is not visible
- integrated catalog api in `Demo.jsx` component in home page
- added toast messages in `Demo.jsx` & `Catalog.jsx` component
- cleaned `ProtectedRoute.jsx` component
- added checks before uploading/updating logo in s3 bucket
- updated failing test-cases in `demo.test.jsx`, `addlogo.js`

- [x] 🍕 Feature
- [x] 👨‍💻 Code Refactor
- [x] ✅ Test

https://github.com/user-attachments/assets/b7d006b2-8903-44fe-bc5b-e797d5fbd0f6

### Test Passed

image

### Screenshots of coverage

image

image

image

## Checklist


packages/app/controllers/catalog.js
```diff
@@ -147,18 +152,10 @@ async function updateCatalogController(req, res, next) {
...
     const Extension = imageName.split(".")[1].toLowerCase();
-    const companyName = Imagename + "." + Extension;
-    if (companyName !== exsitingImage.company_name) {
-      return res.status(400).json({
-        error: STATUS_CODES[400],
-        statusCode: 400,
-        message: Messages.NAME_AND_EXT_SAME,
-      });
-    }
+    const companyName = Imagename;
```
TanishqSingla: what's the purpose of this variable?

packages/app/services/images.js
```diff
@@ -24,14 +24,15 @@ class ImageServices {
     default_extension = "png",
     checkDb = true
   ) {
-    let domainName = company;
+    let domainName = company.company_name;
     if (checkDb) {
-      const image = await this.imageRepository.fetchImage(company);
+      const image = await this.imageRepository.fetchImage(company.company_name);
```
TanishqSingla:
```suggestion
      const image = await this.imageRepository.fetchImage(domainName);
```

packages/app/services/images.js
```diff
@@ -57,7 +58,7 @@ class ImageServices {
     const dataList = [];
     for (const company of companyList) {
```
TanishqSingla: Can we use `Promise.all` or `Promise.allSettled` here?

packages/ui/__tests__/components/Catalog.test.jsx
```diff
@@ -16,13 +83,15 @@ describe("Catalog Component", () => {
 
   it("Add image button should be visible", () => {
     render(<Catalog />);
+    vi.runAllTimers();
```
TanishqSingla: `runAllTimers` is mentioned in `beforeEach`, do we need this after the render?

packages/app/services/images.js
```diff
@@ -24,14 +24,15 @@ class ImageServices {
...
     if (checkDb) {
-      const image = await this.imageRepository.fetchImage(company);
+      const image = await this.imageRepository.fetchImage(company.company_name);
       if (!image) return null;
       domainName = image.company_name;
     }
 
-    const imageUrl = `${default_extension}/${domainName}`;
+    const version = new Date(company.updated_at).getTime();
+    const imageUrl = `${default_extension}/${domainName}?v=${version}`;
```
TanishqSingla: nit: This should be done in a static function containing tests.

packages/app/services/images.js
```diff
@@ -102,15 +103,21 @@ class ImageServices {
   }
 
   async updateImageById(id, updateObj) {
```
TanishqSingla, line 123: please add JSDoc for this

packages/app/services/images.js
```diff
@@ -102,15 +103,21 @@ class ImageServices {
...
+    });
     return {
       _id: updatingImage._id,
       updatedAt: updatingImage.updatedAt,
     };
   }
 
-  async getImagesByUserId(userId) {
-    return await this.imageRepository.getAllImageByUserId(userId);
+  async getImagesByUserId(userId, skip, limit) {
```
TanishqSingla, line 146: Same as above

packages/ui/__tests__/components/Catalog.test.jsx
```diff
@@ -1,11 +1,78 @@
...
+        return Promise.resolve({
+          success: true,
+          message: "Mock operation successful",
+        });
+      });
+
+      return { data, loading, errorMsg, makeRequest };
+    },
+  };
+});
```
TanishqSingla: We should not write logic in tests, this is way too complicated.

packages/ui/__tests__/components/Catalog.test.jsx
```diff
@@ -35,160 +104,187 @@ describe("Catalog Component", () => {
...
     render(<Catalog />);
+    vi.runAllTimers();
 
-    const firstPageCompanies = companies.slice(0, 10);
+    const firstPageCompanies = COMPANIES.slice(0, 10);
     firstPageCompanies.forEach((company) => {
-      const companyImage = screen.getByText(company.companyImage);
-      expect(companyImage).toBeInTheDocument();
+      expect(
+        screen.getByText(
```
TanishqSingla: `getByText` is itself an assertion, so `expect` here is redundant.

packages/ui/__tests__/components/Catalog.test.jsx
```diff
@@ -35,160 +104,187 @@ describe("Catalog Component", () => {
...
+          `${company.company_name.toLowerCase()}.${company.extension}`
+        )
+      ).toBeInTheDocument();
     });
+    if (COMPANIES.length > 10) {
+      expect(
+        screen.queryByText(
+          `${COMPANIES[10].company_name.toLowerCase()}.${COMPANIES[10].extension}`
+        )
+      ).not.toBeInTheDocument();
```
TanishqSingla: `queryByText` returns `null` if the element is not present, so `.not.null()` would be much more accurate

packages/ui/__tests__/components/Catalog.test.jsx
```diff
@@ -35,160 +104,187 @@ describe("Catalog Component", () => {
...
+    vi.runAllTimers();
 
     const searchInput = screen.getByLabelText("search");
     fireEvent.change(searchInput, { target: { value: "Amazon" } });
-    expect(screen.getByText("Amazon.png")).toBeInTheDocument();
-    expect(screen.queryByText("Apple.png")).not.toBeInTheDocument();
+    vi.runAllTimers();
+
+    expect(screen.getByText("amazon.png")).toBeInTheDocument();
+    expect(screen.queryByText("apple.png")).not.toBeInTheDocument();
```
TanishqSingla: Same as above `getBy*` are assertions, so `expect - .toBeInDocument` are redundant

packages/app/repositories/images.js
```diff
@@ -51,8 +51,19 @@ class ImagesRepository extends BaseRepository {
...
   }
 
-  async getAllImageByUserId(userId) {
-    return await this.model.find({ user_id: userId });
+  async getAllImageByUserId(userId, skip, limit) {
+    const total = await this.model.countDocuments({ user_id: userId });
+    const images = await this.model
+      .find({ user_id: userId })
+      .skip(skip)
+      .limit(limit);
```
TanishqSingla: Count query can sometimes be slow, both these queries can be ran in paralllel using `Promise.all`

packages/ui/src/utils/ProtectedRoute.jsx
```diff
@@ -1,54 +1,21 @@
...
-    hasFetched,
-    fetchUserData,
-    isAuthenticated,
-  ]);
 
   if (!isAuthenticated) {
     return <Navigate to="/" replace state={{ from: location.pathname }} />;
   }
 
-  if (!adminOnly) {
```
TanishqSingla, line 30: is removing `adminOnly` part of the issue itself?

packages/ui/src/components/common/modal/Modal.jsx
```diff
@@ -35,8 +35,13 @@ const Modal = ({
...
 
   return (
-    <div className={styles["modal-overlay"]} onClick={handleOverlayClick}>
-      <div
+    <div
+      data-testid="modal-overlay"
+      className={styles["modal-overlay"]}
+      onClick={handleOverlayClick}
+    >
+      <dialog
```
TanishqSingla: html dialog itself renders on a top layer, this will likely render on top of overlay, can you confirm the functionality before is same as after the changes?

packages/ui/src/components/catalog/Catalog.jsx
```diff
@@ -1,53 +1,114 @@
...
-    setSearchTerm("");
-    setPageNum((prevPageNum) => prevPageNum + 1);
   };
 
+  const handleReuploadBtnClick = (id) => {
+    setIsModalOpen(true);
+    setUpdateImageId(id);
+  };
+
+  const filteredCompanies =
```
TanishqSingla: better to have this function wrapped inside `useMemo` so that we don't recompute it every time

packages/ui/src/hooks/useApi.js
```diff
@@ -15,13 +15,15 @@ export const useApi = (config) => {
   const [isSuccess, setIsSuccess] = useState(false);
   const [loading, setLoading] = useState(false);
 
-  const makeRequest = async () => {
+  const makeRequest = async (dynamicConfig = {}) => {
```
TanishqSingla, line 18: Have you evaluated your scenario and compared other components that justifies the need for this?

TanishqSingla (changes requested):

packages/ui/src/hooks/useApi.js
tanwarAalok: image

It would help to pass body data in POST request or update request.

packages/ui/src/utils/ProtectedRoute.jsx
tanwarAalok: We don't need admin logic now in protected route as we are providing dropdown in dashboard itself. There is no additional /admin route. So this part removed.

packages/ui/__tests__/components/Catalog.test.jsx
tanwarAalok: I am not sure about this too, I tried many things before I could make the tests run successfully.

aps08:
Aalok is not contributing, so @TanishqSingla if you are free then you can get started on this.

Allan2000-Git:
@TanishqSingla It would be helpful to add more detailed comments, as the current ones are not very descriptive.

aps08:
@Allan2000-Git @TanishqSingla what is the update on this ?

packages/ui/__tests__/components/Catalog.test.jsx
Allan2000-Git: I have simplified it using the mock data

packages/ui/__tests__/components/Catalog.test.jsx
Allan2000-Git: Yes, not required after render. Have removed all the instances

packages/ui/__tests__/components/Catalog.test.jsx
Allan2000-Git: I have used .not.null() and tested the same

packages/app/services/images.js
Allan2000-Git: have added the jsdoc

packages/app/services/images.js
Allan2000-Git: have added the jsdoc

packages/app/services/images.js
Allan2000-Git: There are no tests containing this line of code

packages/app/services/images.js
```diff
@@ -55,18 +56,27 @@ class ImageServices {
...
-      if (!signedUrl) continue;
-      dataList.push({
-        companyName: company.company_name.split(".")[0],
-        image: signedUrl,
-      });
+    const results = await Promise.allSettled(
+      companyList.map(async (company) => {
+        const signedUrl = await this.fetchImageByCompanyFree(
+          company,
+          undefined,
```
aps08: why passing undefined ?

packages/ui/__tests__/components/Demo.test.jsx
```diff
@@ -37,68 +27,63 @@ describe("Demo Component", () => {
...
-        { companyName: "Amazon", image: "amazon-logo.svg" },
-        { companyName: "Adobe", image: "adobe-logo.svg" },
-        { companyName: "Airbnb", image: "airbnb-logo.svg" },
-      ],
+      data: {
+        data: [
+          { companyName: "Google", image: "google-logo.svg" },
+          { companyName: "Facebook", image: "facebook-logo.svg" },
+          { companyName: "Zoom", image: "zoom-logo.svg" },
+          { companyName: "DoorDash", image: "doordash-logo.svg" },
```
TanishqSingla, line 56: I get the `data` change, but why are the values also changed here, this will just introduce the extra work of changing values in each test

packages/ui/src/utils/Constants.js
```diff
@@ -780,3 +751,10 @@ export const CHANGE_PASSWORD_FIELDS = [
 export const API_KEY_FORM = {
   tagLine: "Generate a new API key to use in your projects.",
 };
+
+export const MOCK_DEMO_COMPANIES = [
+  { companyName: "Google", image: "google-logo.svg" },
+  { companyName: "Facebook", image: "facebook-logo.svg" },
+  { companyName: "Zoom", image: "zoom-logo.svg" },
```
TanishqSingla, line 758: Why do we have mock values in utils? Is this used in tests?

packages/app/services/images.js
```diff
@@ -24,14 +24,15 @@ class ImageServices {
...
     if (checkDb) {
-      const image = await this.imageRepository.fetchImage(company);
+      const image = await this.imageRepository.fetchImage(domainName);
       if (!image) return null;
       domainName = image.company_name;
     }
 
-    const imageUrl = `${default_extension}/${domainName}`;
+    const version = new Date(company.updated_at).getTime();
+    const imageUrl = `${process.env.KEY}/${default_extension}/${domainName}?v=${version}`;
```
TanishqSingla, line 35: This is a method function and you're reading a global variable `process.env.KEY`. Not a good pattern to have predictable code.
Either move this to a function or initialize `KEY` in class in constructor. That way initialized class won't produce different output.

Also can we have a better name than `KEY`, not descriptive enough of where this key is from and where it belongs.

packages/ui/src/components/catalog/Catalog.jsx
```diff
@@ -1,107 +1,205 @@
...
+  } = useApi({
+    method: updateImageId ? "PUT" : "POST",
+    url: `/catalog/logo`,
+    headers: { "Content-Type": "multipart/form-data" },
+  });
+
+  useEffect(() => {
+    makeRequest();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [pageNum]);
```
TanishqSingla, line 43: is the interface going to be an infinite scroll pagination? If not then why are we not having support for url search params which will allow the user to have browser navigation controls

TanishqSingla (changes requested):

PR #799, opened 2025-08-13
fix: logo not found in business api

### LINKED ISSUE #784: Logo not found error in Business logo API
### Describe your issue

- The `api/logo` endpoint giving status `404` when quering for a logo that already exists 
- when we try to search the same logo in `api/logo/search` endpoint. we get the image url with permission denied.

### zoom logo query 

Image

### zoom logo search 

Image

### Confirmation

theboyofdream:
## Description

Closes Issue: #784 

- Fix logo API image fetch by passing object instead of string

Extra
- Updated api-stage-openlogo.fyi link constant file to api-stage.openlogo.fyi
- Updated jsDoc inside app/services --file-> images, subscription.

- [x] 🐛 Bug Fix
- [x] 📄 Documentation Update

### Logo Retrieval API

image

### Logo Search API 

image

All test-cases passed locally

image

## Checklist


Aniketvish0 (dismissed):

packages/app/controllers/logo.js
```diff
@@ -51,7 +51,9 @@ async function getLogoController(req, res, next) {
       });
     }
 
-    const imageUrl = await imageService.fetchImageByCompanyFree(company);
+    const imageUrl = await imageService.fetchImageByCompanyFree({
```
TanishqSingla: This function doesn't take object as arguments, why are we passing company_name like that?

TanishqSingla (changes requested):

TanishqSingla:
Also why were tests not changed, this is a complete function signature change. Don't we have adequate testing for this?

packages/app/controllers/logo.js
theboyofdream: image

inside function we are using `company.company_name` so function expects an object

also now after you pointed out 
1. i see that function jsdoc is wrong which i will update
2. we are using `default_extension` variable name which is inconsistent so i am thinking of renaming it to `defaultExtension`
3. this func currently used in 2 places - should i use companyName <string> instead of company <Object>

theboyofdream: > This function doesn't take object as arguments, why are we passing company_name like that?

image

inside function we are using `company.company_name` so function expects an object

also now after you pointed out 
1. i see that function jsdoc is wrong which i will update
2. we are using `default_extension` variable name which is inconsistent so i am thinking of renaming it to `defaultExtension`
3. this func currently used in 2 places - should i use `companyName <string>` instead of `company <Object>`

TanishqSingla: ok, fix jsdoc then

packages/app/services/subscriptions.js
```diff
@@ -9,15 +9,15 @@ class SubscriptionService {
   /**
    * Gets Subscription by Id.
    * @param {string} subscriptionId - The subscription_id of the user.
-   * @returns {Object} - Subscription Object.
+   * @returns {Promise<Object>} - Subscription Object.
```
TanishqSingla, line 12: nit: remove the type here, the lsp can infer

TanishqSingla (approved):

aps08:
Will deploy it in the evening. @theboyofdream

PR #826, opened 2025-09-06
fix(usage-count): prevent increment when logo fetch returns 404

### LINKED ISSUE #823: `v0.3.0` Usage count increments even when logo not found (404)
### Describe your issue

When calling the GET /logo route, the monthly usage count is incremented even if the response is 404 Not Found.

This leads to:-

- Users being charged quota for failed fetch attempts.
- Quota being consumed without delivering any usable result.

### Steps to reproduce

Steps to Reproduce
1. Log in to the app.
2. Generate an API key.
3. Call the GET /logo endpoint with a valid API key, but use a company name that does not exist (e.g., /logo?key=nonexistent&API_KEY=YOUR_API_KEY).
4. Observe that the response is 404 Not Found, but the subscription usage_count is still incremented.

https://github.com/user-attachments/assets/fb6c1f52-171b-49b6-95da-82e6d63779e8

### Confirmation

MethodManav:
## Description
Fixes #823 
This PR updates the GET /logo route so that the monthly usage count is only incremented when the response is successful (200). Previously, the count was incremented even when a 404 was returned, which incorrectly consumed the user’s quota for failed fetch attempts.

- [x] 🐛 Bug Fix
- [ ] 👨‍💻 Code Refactor
- [ ]  🍕 Feature
- [ ] 🐛 Bug Fix
- [ ]  📄 Documentation Update
- [ ]  👨‍💻 Code Refactor
- [ ]  🔥 Performance Improvements
- [ ]  ✅ Test
- [ ]  🛠️ CI/CD

https://github.com/user-attachments/assets/bb1166f5-dcbf-4234-88f0-d7f3eed2ee5e

## Checklist

### DIFF
--- a/packages/app/controllers/logo.js
+++ b/packages/app/controllers/logo.js
@@ -59,7 +59,15 @@ async function getLogoController(req, res, next) {
         error: STATUS_CODES[404],
       });
     }
-    await subscriptionService.incrementUsageCount(userSubscription);
+    const updatedUsageCount =
+      await subscriptionService.incrementUsageCount(userSubscription);
+    if (!updatedUsageCount) {
+      return res.status(500).json({
+        message: Messages.INTERNAL_SERVER_ERROR,
+        statusCode: 500,
+        error: STATUS_CODES[500],
+      });
+    }
     return res.status(200).json({
       statusCode: 200,
       data: imageUrl,
--- a/packages/app/controllers/logo.js
+++ b/packages/app/controllers/logo.js
@@ -52,7 +52,6 @@ async function getLogoController(req, res, next) {
     }
 
     const imageUrl = await imageService.fetchImageByCompanyFree(company);
-    await subscriptionService.incrementUsageCount(userSubscription);
     if (!imageUrl) {
       return res.status(404).json({
         message: Messages.LOGO_NOT_FOUND,
@@ -60,7 +59,7 @@ async function getLogoController(req, res, next) {
         error: STATUS_CODES[404],
       });
     }
-
+    await subscriptionService.incrementUsageCount(userSubscription);
     return res.status(200).json({
       statusCode: 200,
       data: imageUrl,
--- a/packages/app/__tests__/controllers/logo/get-logo.js
+++ b/packages/app/__tests__/controllers/logo/get-logo.js
@@ -138,4 +138,24 @@ describe("getLogoController", () => {
     const response = await request(app).get(apiUrl).query(baseQuery);
     expect(response.status).toBe(500);
   });
+
+  it("should handle unexpected errors if usage count is null or undefined", async () => {
+    mockRepetedService(mockSubscription[0]);
+
+    jest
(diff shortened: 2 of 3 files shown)


Hariom01010 (dismissed):

packages/app/controllers/logo.js
```diff
@@ -52,15 +52,14 @@ async function getLogoController(req, res, next) {
...
-    await subscriptionService.incrementUsageCount(userSubscription);
     if (!imageUrl) {
       return res.status(404).json({
         message: Messages.LOGO_NOT_FOUND,
         statusCode: 404,
         error: STATUS_CODES[404],
       });
     }
-
+    await subscriptionService.incrementUsageCount(userSubscription);
```
TanishqSingla: This can be an async operation why are we making it block the api call?

TanishqSingla (changes requested):

packages/app/controllers/logo.js
MethodManav: I feel it’s safer to await the DB update here — otherwise if the update fails after the response is sent, the usage count could become inconsistent. Let me know if you’d prefer optimizing for speed instead, but since quota tracking is critical, I lean towards keeping the await

TanishqSingla: Since the issue is unrelated I think it's no use discussing it here now, but we should come up with a better solution than doing this.

About the safety issue I don't see any error handling on incrementUsageCount call, what are the scenarios it will fail?
How are we handling those failure, what if incrementUsageCount only returns null on error or a object

TanishqSingla (approved):

packages/app/controllers/logo.js
```diff
@@ -52,15 +52,22 @@ async function getLogoController(req, res, next) {
...
         statusCode: 404,
         error: STATUS_CODES[404],
       });
     }
-
+    const updatedUsageCount =
+      await subscriptionService.incrementUsageCount(userSubscription);
+    if (!updatedUsageCount) {
+      return res.status(500).json({
+        message: Messages.INTERNAL_SERVER_ERROR,
```
aps08, line 66: don't throw error here. just log the error message.

packages/app/controllers/logo.js
```diff
@@ -52,15 +52,22 @@ async function getLogoController(req, res, next) {
...
-    await subscriptionService.incrementUsageCount(userSubscription);
     if (!imageUrl) {
       return res.status(404).json({
         message: Messages.LOGO_NOT_FOUND,
         statusCode: 404,
         error: STATUS_CODES[404],
       });
     }
-
+    const updatedUsageCount =
```
aps08, line 62: remove `const updatedUsageCount =`

parshant0717:
Hey, I've taken over this PR and added the necessary changes. I'll be creating a new pull request from my branch (bugfix/usage-count) to finalize this. Closing this PR in favor of the new one.

PR #833, opened 2025-09-08
Schedule every day run of refresh api script

### LINKED ISSUE #825: `v0.3.0` Schedule everyday run of refresh api limit script
### Describe the enhancement

- currently there is action written already in the `.github/workflows` folder.
- the actual limit refresh file is yet to be authored.
- for all users the api usage limit should refresh on 1st of every month. 

- this action should run daily to check if it is 1st of the month.
- on 1st of the month, for every user the api limit should reset to zero(0).

### Confirmation

Aniketvish0:
## Description
closes #825

- added the script to refresh the usage api limit to 0.
- updated the workflow, added cron to run the workflow at midnight everyday
- runs on both prod and stage branch parallely
- added logs to show on action logs to identify that the scripts is running fine.
- added the test for unit testing.
- this script uses {MONGO_URL} environment variable

- [x] 🍕 Feature
- [x] ✅ Test
- [x] 🛠️ CI/CD

## Screenshots 
- scripts runs on schedule
Screenshot from 2025-09-16 14-04-23
Screenshot from 2025-09-16 14-04-42

## Checklist


.github/workflows/refresh-api-limit-daily.yaml
```diff
@@ -1,6 +1,8 @@
 name: Refresh API limit Daily
 
 on:
```
aps08, line 3: on which branch it this going to work ?

packages/app/scripts/refreshUsageCount.js
```diff
@@ -0,0 +1,46 @@
+const mongoose = require("mongoose");
+
+const Subscriptions = require("../models/subscriptions");
+
+async function refreshUsageCount() {
+  try {
+    const today = new Date();
+    const isFirstDayOfMonth = today.getDate() === 1;
+    if (!isFirstDayOfMonth) {
+      console.log("Not the first day of the month, skipping...");
```
aps08: why do we need this ?

aps08:
PR doesn't mentioned about any environment variable getting used but the code is using.

.github/workflows/refresh-api-limit-daily.yaml
Aniketvish0: updated this on line 10

Aniketvish0:
> PR doesn't mentioned about any environment variable getting used but the code is using.

updated the description

aps08 (changes requested):
Should use `environment`.
Should run on prod also.
Correct branch selection, avoid using `if`

Aniketvish0:
@aps08 I have added the environments, now it will run on both prod and stage branch and will use the secrets according to environments, but the action will fail on the prod branch as the prod will not consist the script as of now, so we need to merge this with prod branch too.

theboyofdream (dismissed):

PR #843, opened 2025-09-17
feat(admin): enable global search across database

### LINKED ISSUE #831: `v0.3.0` Make search in Admin Dashboard work across the entire database
### Describe the enhancement

Currently, the search option in the Admin Dashboard only filters results on the client-side within the current page of data.

The enhancement should update the API and frontend so that the search query is sent to the backend, and filtering is done at the database level. This way, search results will include all matching records from the entire database, not just the currently loaded page.

This will let admins find logos/companies across the whole system, not just within one page.

### Confirmation

biplab-sutradhar:
## Description

closes #831 

This PR makes the Admin Dashboard search work across the **entire database** instead of just filtering the current page.

* Added backend support for `search` query.
* Updated frontend to send search term to API.
* Added tests for matches and empty results.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🛠️ CI/CD

before: 
image

after:

https://github.com/user-attachments/assets/89ab30ae-3ba2-4e77-8b49-7f31ed71a7cd

## Checklist

### DIFF
--- a/packages/app/repositories/images.js
+++ b/packages/app/repositories/images.js
@@ -52,7 +52,7 @@ class ImagesRepository extends BaseRepository {
   }
 
   async getAllImageByUserId(userId, skip, limit, search = "") {
-    const query = { user_id: userId };
+    const query = {};
 
     if (search) {
       query.company_name = { $regex: search, $options: "i" };
--- a/packages/app/repositories/images.js
+++ b/packages/app/repositories/images.js
@@ -51,10 +51,16 @@ class ImagesRepository extends BaseRepository {
     return cloudFrontUrl;
   }
 
-  async getAllImageByUserId(userId, skip, limit) {
+  async getAllImageByUserId(userId, skip, limit, search = "") {
+    const query = { user_id: userId };
+
+    if (search) {
+      query.company_name = { $regex: search, $options: "i" };
+    }
+
     const [total, images] = await Promise.all([
-      this.model.countDocuments({ user_id: userId }),
-      this.model.find({ user_id: userId }).skip(skip).limit(limit),
+      this.model.countDocuments(query),
+      this.model.find(query).skip(skip).limit(limit).sort({ createdAt: -1 }),
     ]);
 
     return {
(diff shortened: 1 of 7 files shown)


packages/ui/__tests__/components/Catalog.test.jsx
AryaDharkar, line 1: vi.runAllTimers() in beforeEach is unnecessary (no timers pending before a test starts).
Just keep vi.clearAllMocks() in beforeEach and vi.clearAllTimers() in afterEach.

AryaDharkar (changes requested):

packages/app/repositories/images.js
```diff
@@ -51,10 +51,16 @@ class ImagesRepository extends BaseRepository {
     return cloudFrontUrl;
   }
 
-  async getAllImageByUserId(userId, skip, limit) {
+  async getAllImageByUserId(userId, skip, limit, search = "") {
+    const query = {};
```
aps08, line 55: is this needed here if it's empty ?

packages/app/services/images.js
```diff
@@ -143,8 +143,13 @@ class ImageServices {
    * @param {number} limit - The maximum number of images to return.
    * @throws {Error} If the retrieval operation fails.
    */
-  async getImagesByUserId(userId, skip, limit) {
-    return await this.imageRepository.getAllImageByUserId(userId, skip, limit);
+  async getImagesByUserId(userId, skip, limit, search = "") {
```
aps08: I believe this function name should change. If userid is not a constraint

packages/app/repositories/images.js
biplab-sutradhar: Yes, it is needed because values will be added to it later and it will be passed into the database query

theboyofdream (commented):
1. can we implement the deboucing for search
2. also search string should be at least 2 chars

theboyofdream (approved):

PR #849, opened 2025-09-24
Deletion tenure to prevent immediate re-registration

### LINKED ISSUE #848: `v0.3.0` Prevent Immediate Re-Signup for Deleted Users
### Describe the enhancement

This enhancement implements a soft delete for users and restricts re-registration for a configurable tenure period.
* When a user deletes their account → mark deletedAt and blockUntil = deletedAt + X days (tenure period configurable).
* After the block period → allow re-registration or account reactivation.

Currently, deleted users can potentially re-register immediately, which may lead to abuse or confusion. Implementing a soft delete with a configurable block period:
* Prevents repeated delete/re-signup abuse.
* Provides a cooling-off period for users.
* Keeps normal users’ login/signup flow unaffected.
* The block period can be modified according to business needs, e.g., ensuring paid/free plans tied to this account are respected before allowing re-registration.

### Confirmation

suchit-10:
## Description

Currently, deleted users can potentially re-register immediately, which may lead to abuse or confusion. Implementing a soft delete with a configurable block period:

Closes: #848

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🛠️ CI/CD

https://github.com/user-attachments/assets/74d09d59-9bc7-4577-b6bf-9db1388750e7

## Checklist

### DIFF
--- a/packages/app/__tests__/controllers/auth/signin.js
+++ b/packages/app/__tests__/controllers/auth/signin.js
@@ -45,9 +45,7 @@ describe("SIGNIN API", () => {
   });
 
   it("404 - Incorrect email or password", async () => {
-    jest
-      .spyOn(UserService.prototype, "getAnyUserByEmail")
-      .mockResolvedValue(null);
+    jest.spyOn(UserService.prototype, "getUserByEmail").mockResolvedValue(null);
 
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
@@ -69,9 +67,7 @@ describe("SIGNIN API", () => {
       generateJWT: jest.fn().mockReturnValue("jwt-token"),
     };
 
-    jest
-      .spyOn(UserService.prototype, "getAnyUserByEmail")
-      .mockResolvedValue(user);
+    jest.spyOn(UserService.prototype, "getUserByEmail").mockResolvedValue(user);
 
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
@@ -94,9 +90,7 @@ describe("SIGNIN API", () => {
       generateJWT: jest.fn().mockReturnValue("jwt-token"),
     };
 
-    jest
-      .spyOn(UserService.prototype, "getAnyUserByEmail")
-      .mockResolvedValue(user);
+    jest.spyOn(UserService.prototype, "getUserByEmail").mockResolvedValue(user);
 
     const INVAILD = "wrongpassword";
     const response = await request(app)
@@ -120,9 +114,7 @@ describe("SIGNIN API", () => {
       generateJWT: jest.fn().mockReturnValue("jwt-token"),
     };
 
-    jest
-      .spyOn(UserService.prototype, "getAnyUserByEmail")
-      .mockResolvedValue(user);
+    jest.spyOn(UserService.prototype, "getUserByEmail").mockResolvedValue(user);
 
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
--- a/packages/app/__tests__/controllers/auth/signin.js
+++ b/packages/app/__tests__/controllers/auth/signin.js
@@ -71,9 +71,10 @@ describe("SIGNIN API", () => {
       .spyOn(UserService.prototype, "getAnyUserByEmail")
       .mockResolvedValue(user);
 
+    const TESTPASSWORD = "password123";
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
-      .send({ email: user.email, password: "password123" });
+      .send({ email: user.email, password: TESTPASSWORD });
 
     expect(response.status).toBe(403);
     expect(response.body).toEqual({
@@ -96,9 +97,10 @@ describe("SIGNIN API", () => {
       .spyOn(UserService.prototype, "getAnyUserByEmail")
       .mockResolvedValue(user);
 
+    const INVAILD = "wrongpassword";
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
-      .send({ email: user.email, password: "wrongpassword" });
+      .send({ email: user.email, password: INVAILD });
 
     expect(response.status).toBe(404);
     expect(response.body).toEqual({
@@ -121,9 +123,10 @@ describe("SIGNIN API", () => {
       .spyOn(UserService.prototype, "getAnyUserByEmail")
       .mockResolvedValue(user);
 
+    const TESTPASSWORD = "password123";
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
-      .send({ email: user.email, password: "password123" });
+      .send({ email: user.email, password: TESTPASSWORD });
 
     expect(response.status).toBe(200);
     expect(response.body).toEqual({ statusCode: 200 });
--- a/packages/app/__tests__/controllers/auth/signin.js
+++ b/packages/app/__tests__/controllers/auth/signin.js
@@ -73,7 +73,6 @@ describe("SIGNIN API", () => {
       .spyOn(UserService.prototype, "getAnyUserByEmail")
       .mockResolvedValue(user);
 
-    // const TESTPASSWORD = "password123";
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
       .send({ email: user.email, password: dummyPassword });
@@ -125,7 +124,6 @@ describe("SIGNIN API", () => {
       .spyOn(UserService.prototype, "getAnyUserByEmail")
       .mockResolvedValue(user);
 
-    // const TESTPASSWORD = "password123";
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
       .send({ email: user.email, password: dummyPassword });
--- a/packages/app/__tests__/controllers/auth/signin.js
+++ b/packages/app/__tests__/controllers/auth/signin.js
@@ -6,6 +6,8 @@ const { ENDPOINTS } = require("../../../utils/testconstants");
 const { MOCK_USERS } = require("../../../utils/mocks");
 const { Messages } = require("../../../utils/constants");
 const app = require("../../../server");
+const dummyPassword =
+  require("../../../utils/generatePassword").generatePassword();
 
 describe("SIGNIN API", () => {
   beforeAll(() => {
@@ -71,10 +73,10 @@ describe("SIGNIN API", () => {
       .spyOn(UserService.prototype, "getAnyUserByEmail")
       .mockResolvedValue(user);
 
-    const TESTPASSWORD = "password123";
+    // const TESTPASSWORD = "password123";
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
-      .send({ email: user.email, password: TESTPASSWORD });
+      .send({ email: user.email, password: dummyPassword });
 
     expect(response.status).toBe(403);
     expect(response.body).toEqual({
@@ -123,10 +125,10 @@ describe("SIGNIN API", () => {
       .spyOn(UserService.prototype, "getAnyUserByEmail")
       .mockResolvedValue(user);
 
-    const TESTPASSWORD = "password123";
+    // const TESTPASSWORD = "password123";
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
-      .send({ email: user.email, password: TESTPASSWORD });
+      .send({ email: user.email, password: dummyPassword });
 
     expect(response.status).toBe(200);
     expect(response.body).toEqual({ statusCode: 200 });
--- a/packages/app/__tests__/controllers/auth/signin.js
+++ b/packages/app/__tests__/controllers/auth/signin.js
@@ -43,7 +43,10 @@ describe("SIGNIN API", () => {
   });
 
   it("404 - Incorrect email or password", async () => {
-    jest.spyOn(UserService.prototype, "getUserByEmail").mockResolvedValue(null);
+    jest
+      .spyOn(UserService.prototype, "getAnyUserByEmail")
+      .mockResolvedValue(null);
+
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
       .send({ email: "email@example.com", password: "password" });
@@ -57,28 +60,45 @@ describe("SIGNIN API", () => {
   });
 
   it("403 - Email not verified", async () => {
+    const user = {
+      ...MOCK_USERS[0],
+      is_verified: false,
+      matchPassword: jest.fn().mockResolvedValue(true),
+      generateJWT: jest.fn().mockReturnValue("jwt-token"),
+    };
+
     jest
-      .spyOn(UserService.prototype, "getUserByEmail")
-      .mockResolvedValue(MOCK_USERS[0]);
+      .spyOn(UserService.prototype, "getAnyUserByEmail")
+      .mockResolvedValue(user);
+
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
-      .send({ email: "johndoe@example.com", password: "password123" });
+      .send({ email: user.email, password: "password123" });
 
     expect(response.status).toBe(403);
     expect(response.body).toEqual({
       error: STATUS_CODES[403],
-      message: "Email not verified",
+      message: Messages.EMAIL_NOT_VERIFIED,
       statusCode: 403,
     });
   });
 
-  it("404 - Incorrect email or password", async () => {
+  it("404 - Incorrect password", async () => {
+    const user = {
+      ...MOCK_USERS[1],
+      is_verified: true,
+      is_deleted: false,
+      matchPassword: jest.fn().mockResolvedValue(false),
+      generateJWT: jest.fn().mockReturnValue("jwt-token"),
+    };
+
     jest
-      .spyOn(UserService.prototype, "getUserByEmail")
-      .mockImplementation(() => new Users(MOCK_USERS[1]));
+      .spyOn(UserService.prototype, "getAnyUserByEmail")
+      .mockResolvedValue(user);
+
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
-      .send({ email: "johndoe1@example.com", password: "password122" });
+      .send({ email: user.email, password: "wrongpassword" });
 
     expect(response.status).toBe(404);
     expect(response.body).toEqual({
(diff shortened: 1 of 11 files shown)


packages/app/__tests__/controllers/auth/signin.js
```diff
@@ -57,28 +62,46 @@ describe("SIGNIN API", () => {
...
     const response = await request(app)
       .post(ENDPOINTS.SIGNIN)
-      .send({ email: "johndoe@example.com", password: "password123" });
+      .send({ email: user.email, password: dummyPassword });
 
     expect(response.status).toBe(403);
     expect(response.body).toEqual({
       error: STATUS_CODES[403],
-      message: "Email not verified",
+      message: Messages.EMAIL_NOT_VERIFIED,
```
TanishqSingla, line 79: Please don't change test cases unless they are breaking, also try to refrain from using constants or object values that are dynamic.

packages/app/models/users.js
```diff
@@ -42,10 +42,21 @@ const userSchema = new mongoose.Schema({
...
   is_deleted: {
     type: Boolean,
     default: false,
   },
+
+  deleted_at: {
+    type: Date,
+    default: null,
+  },
+  block_until: {
```
TanishqSingla: Instead of having a new field for block duration, we can just use `deleted_at` and `is_deleted` date and programmatically add checks,

packages/app/services/users.js
```diff
@@ -103,6 +125,12 @@ class UserService {
   async getUserByEmail(email) {
     return await this.userRepository.findUserByEmail(email);
   }
+  /**
+   * Fetches a user by email, including deleted users.
+   */
+  async getAnyUserByEmail(email) {
+    return await this.userRepository.findAnyUserByEmail(email);
+  }
```
TanishqSingla: The previous function literally does this. Please don't implement features without looking at code.

packages/app/services/users.js
```diff
@@ -91,10 +91,32 @@ class UserService {
...
+    if (user.block_until && user.block_until > new Date()) {
+      const daysLeft = Math.ceil(
+        (user.block_until - new Date()) / (1000 * 60 * 60 * 24)
+      );
+      throw new Error(
+        `Account exists. You may re-register after ${daysLeft} days.`
+      );
+    }
+    return true;
+  }
```
TanishqSingla: I don't think a db call should be required just to return a boolean value, this can be done at controller level.

packages/app/repositories/users.js
```diff
@@ -21,6 +21,15 @@ class UsersRepository extends BaseRepository {
     return await this.model.findOne({ email: emailId, is_deleted: false });
   }
 
+  async findAnyUserByEmail(email) {
+    return await this.model.findOne({ email });
+  }
```
TanishqSingla: Why this function is created in 2 files?

packages/app/repositories/base.js
```diff
@@ -37,15 +37,27 @@ class BaseRepository {
   }
 
   async update(id, data) {
-    return await this.model.findByIdAndUpdate(id, data);
+    return await this.model.findByIdAndUpdate(id, data, { new: true });
```
TanishqSingla: Please confirm this doesn't break any existing functions using this.

packages/app/controllers/auth.js
```diff
@@ -140,9 +149,9 @@ async function signinController(req, res, next) {
     /** @type {import("express").CookieOptions}  */
     const cookieOptions = {
       expires: oneDayValidityTimestamp,
-      sameSite: "strict",
-      httpOnly: true,
-      domain: ".openlogo.fyi",
+      // sameSite: "strict",
+      // httpOnly: true,
+      // domain: ".openlogo.fyi",
```
TanishqSingla: Again please check your PR before requesting for review.

packages/app/controllers/auth.js
```diff
@@ -173,9 +182,9 @@ function signoutController(req, res, next) {
 
     /** @type {import("express").CookieOptions}  */
     const cookieOptions = {
-      sameSite: "strict",
-      httpOnly: true,
-      domain: ".openlogo.fyi",
+      // sameSite: "strict",
+      // httpOnly: true,
+      // domain: ".openlogo.fyi",
```
TanishqSingla: Again why this exists in your PR?

packages/app/repositories/base.js
```diff
@@ -37,15 +37,27 @@ class BaseRepository {
...
+  async softDelete(id, tenureDays = 30) {
+    const blockUntil = new Date();
+    blockUntil.setDate(blockUntil.getDate() + tenureDays);
+
+    return await this.model.findByIdAndUpdate(
+      id,
+      {
+        is_deleted: true,
+        deleted_at: new Date(),
+        block_until: blockUntil,
```
TanishqSingla: don't need `block_until` handle this logic at controller level, that way it can be better tested.

packages/app/repositories/base.js
```diff
@@ -37,15 +37,27 @@ class BaseRepository {
...
   }
 
   async delete(id) {
     return await this.model.findByIdAndDelete(id);
   }
 
-  async mark_deleted(id) {
-    return await this.model.findByIdAndUpdate(id, { isDeleted: true });
+  //Tenure-based soft delete
+  async softDelete(id, tenureDays = 30) {
```
TanishqSingla: This function is currenlty only used for records that need to be flagged as delete, and it is not generic, it is better to have them in inherited class.

packages/app/repositories/users.js
TanishqSingla:
```suggestion
  async findUserByEmail(emailId, isDeleted = false) {
```

Use this function instead of creating a new function that does a similar thing

TanishqSingla (changes requested):
Recommended some changes, also please include test coverage reports.

packages/app/services/users.js
suchit-10: Fixed with exisiting function `findUserByEmail`

packages/app/services/users.js
suchit-10: Moved into controller with removed `block_until`

packages/app/repositories/users.js
suchit-10: Proceed with exisiting `findUserByEmail`

packages/app/controllers/auth.js
suchit-10: Uncomment in latest PR.

packages/app/repositories/base.js
suchit-10: Removed `block_until` and replaced the logic with `deleted_at` and `is_deleted` and moved to the controller.

packages/app/repositories/users.js
suchit-10: Used 'findUserByEmail` for write the logic.

suchit-10: Used exisiting `findUserByEmail` to write a logic.

packages/app/controllers/auth.js
suchit-10: Uncomment in latest PR

packages/app/repositories/base.js
suchit-10: Removed the `softDelete` function used the exisiting `delete` func

packages/app/repositories/base.js
suchit-10: I have verified usages of this update method. returning the updated document `({ new: true })` is consistent with our API this change will not break existing functions

packages/app/__tests__/controllers/auth/signin.js
suchit-10: It is already written as in contants.js file

packages/app/controllers/auth.js
AryaDharkar, line 1: Do not use date object so many times. Just call it once in the beginning. Also add the math into a variable for easier understanding. Something like-

  const MS_PER_DAY = 1000 * 60 * 60 * 24;
  const expiry = deletedAt + TENURE_DAYS * MS_PER_DAY;
  const now = Date.now();

packages/app/services/users.js
```diff
@@ -103,7 +103,9 @@ class UserService {
   async getUserByEmail(email) {
     return await this.userRepository.findUserByEmail(email);
   }
-
+  /**
+   * Fetches a user by email, including deleted users.
+   */
```
TanishqSingla: The function name and comment doesn't  match

packages/app/repositories/users.js
```diff
@@ -17,8 +17,13 @@ class UsersRepository extends BaseRepository {
...
    * @returns {Promise<Object|null>} - Returns the user document if found, otherwise null.
    */
-  async findUserByEmail(emailId) {
-    return await this.model.findOne({ email: emailId, is_deleted: false });
+  async findUserByEmail(email) {
+    return await this.model.findOne({ email });
+  }
+
+  //delete user with tenure
+  delete(userId) {
```
TanishqSingla: The comment and function don't match up, please verify this

AryaDharkar (changes requested):

packages/app/models/users.js
```diff
@@ -42,10 +42,16 @@ const userSchema = new mongoose.Schema({
     type: [mongoose.Schema.Types.ObjectId],
     ref: "keys",
   },
+
```
TanishqSingla: don't add gaps like this, if they are not a convention

packages/app/controllers/auth.js
TanishqSingla: We're already using dayjs, library. It will be better to utilize that library instead.

packages/app/controllers/auth.js
```diff
@@ -33,15 +34,39 @@ async function signupController(req, res, next) {
...
 
+    // Handle soft-deleted users
+    if (user?.deleted_at) {
+      const tenureDays = 30;
+      const expiry = new Date(user.deleted_at);
+      expiry.setDate(expiry.getDate() + tenureDays);
+
+      if (expiry > new Date()) {
+        const daysLeft = Math.ceil(
+          (expiry - Date.now()) / (1000 * 60 * 60 * 24)
```
TanishqSingla: Use dayjs library for arithmetic involving date

packages/app/controllers/auth.js
```diff
@@ -33,15 +34,39 @@ async function signupController(req, res, next) {
...
     const { email } = value;
-    const emailExists = await userService.getUserByEmail(email);
-    if (emailExists) {
+    let user = await userService.getUserByEmail(email);
+
+    // If the user is active (not deleted), block signup
+    if (user?.is_deleted === false) {
       return res.status(400).json({
-        message: Messages.EMAIL_EXISTS,
+        message: "Account already exists.",
```
TanishqSingla: Why are you not using constants?

packages/app/controllers/auth.js
```diff
@@ -33,15 +34,39 @@ async function signupController(req, res, next) {
...
+      const tenureDays = 30;
+      const expiry = new Date(user.deleted_at);
+      expiry.setDate(expiry.getDate() + tenureDays);
+
+      if (expiry > new Date()) {
+        const daysLeft = Math.ceil(
+          (expiry - Date.now()) / (1000 * 60 * 60 * 24)
+        );
+        return res.status(400).json({
+          message: `Account exists. You may re-register after ${daysLeft} days.`,
```
TanishqSingla, line 55: For text like this make sure to add them in test case.

packages/app/controllers/auth.js
```diff
@@ -79,7 +105,10 @@ async function signupController(req, res, next) {
       },
     });
 
-    return res.status(200).json({ statusCode: 200 });
+    return res.status(201).json({
+      message: "User created successfully. Please verify your email.",
```
TanishqSingla: Create a constant for static strings or add these static strings in tests.

packages/app/controllers/auth.js
```diff
@@ -117,12 +146,20 @@ async function signinController(req, res, next) {
...
       if (!user.is_verified)
         return res.status(403).json({
           error: STATUS_CODES[403],
           message: Messages.EMAIL_NOT_VERIFIED,
           statusCode: 403,
         });
+      if (user.is_deleted) {
+        return res.status(400).json({
+          error: STATUS_CODES[400],
+          message: "Account does not exist.",
```
TanishqSingla: Same as above

TanishqSingla, line 159: How does this message sounds?

```suggestion
          message: "Email not available.",
```

packages/app/controllers/auth.js
```diff
@@ -53,6 +78,7 @@ async function signupController(req, res, next) {
     Object.assign(value, { subscription_id: newSubscription._id });
     const newUser = await userService.createUser(value);
     if (!newUser) {
+      await subscriptionService.deleteSubscription(newSubscription._id);
```
TanishqSingla: Why are we deleting subscription record and throwing error

TanishqSingla (changes requested):

packages/app/controllers/auth.js
suchit-10: Create subscription only after validation passes hence no use.

packages/app/controllers/auth.js
suchit-10: Fixed with 'USER_CREATED'

suchit-10: Fixed with  `USER_CREATED` in constant

packages/app/controllers/auth.js
suchit-10: Fixed

packages/app/controllers/auth.js
suchit-10: Implemented using dayjs lib.

packages/app/models/users.js
suchit-10: Removed.

packages/app/services/users.js
suchit-10: removed use existing service instead.

packages/app/repositories/users.js
suchit-10: Fixed.

packages/app/controllers/auth.js
suchit-10: It is true.

packages/app/controllers/auth.js
suchit-10: Test case added.

packages/app/controllers/auth.js
```diff
@@ -33,15 +34,33 @@ async function signupController(req, res, next) {
     }
 
     const { email } = value;
-    const emailExists = await userService.getUserByEmail(email);
-    if (emailExists) {
+    let user = await userService.getUserByEmail(email);
```
TanishqSingla:
```suggestion
    const user = await userService.getUserByEmail(email);
```

packages/app/controllers/auth.js
TanishqSingla: This is regarding constants insteaed of hardcoded strings.

packages/app/services/users.js
```diff
@@ -85,14 +86,43 @@ class UserService {
     return destroyedKey;
   }
 
-  /**
+  /*
+   * Soft delete a user: marks as deleted and adds deleted_at timestamp.
```
TanishqSingla:
```suggestion
   * Marks `is_deleted` attribute of the user to true and adds deleted_at timestamp.
```

packages/app/services/users.js
```diff
@@ -85,14 +86,43 @@ class UserService {
...
+    const TENURE_DAYS = 30;
+    if (user?.is_deleted && user?.deleted_at) {
+      const expiry = dayjs(user.deleted_at).add(TENURE_DAYS, "day");
+
+      if (expiry.isBefore(dayjs())) {
+        const deletedUser = await this.userRepository.delete(userId);
+        return !!deletedUser;
+      }
+    }
+
```
TanishqSingla: Most of the logic should be handled at controller level. Services should serve the purpose of retrieving data and transforming them for easier use.

packages/app/services/users.js
```diff
@@ -103,7 +133,10 @@ class UserService {
   async getUserByEmail(email) {
     return await this.userRepository.findUserByEmail(email);
   }
-
+  /**
+   * Fetch a user by email, including soft-deleted users.
```
TanishqSingla: I don't see any parameter passed for email also do changes on your functions only. Don't touch other parts of code that are not required.

packages/app/__tests__/controllers/auth/signup.js
```diff
@@ -258,19 +293,107 @@ describe("SIGNUP API", () => {
...
       .post(ENDPOINTS.SIGNUP)
       .send(mockRequest);
 
-    expect(response.status).toBe(200);
-    expect(response.body.statusCode).toBe(200);
+    const expiry = dayjs(deletedUser.deleted_at).add(30, "day");
+    const daysLeft = Math.ceil(expiry.diff(dayjs(), "day", true));
+
+    expect(response.body.message).toBe(
+      `Account exists. You may re-register after ${daysLeft} day${daysLeft > 1 ? "s" : ""}.`
```
TanishqSingla: Don't use variables for assertions

TanishqSingla (changes requested):

packages/app/__tests__/controllers/auth/signup.js
suchit-10: Ok,
I used it because 30 days then 29,.. 2 days then "1 day"not days.

packages/app/controllers/auth.js
suchit-10: Fixed and tested also.

packages/app/services/users.js
suchit-10: Removed.

packages/app/__tests__/controllers/auth/signup.js
suchit-10: Fixed removed assertions.

packages/app/services/users.js
suchit-10: Done.

packages/app/__tests__/controllers/auth/signup.js
TanishqSingla: The test string should not be dynamic it should be static

TanishqSingla (approved):

PR #876, opened 2025-11-14
Backend: resend verification email logic and tests

### LINKED ISSUE #766: `0.5.0` Add "Resend Verification Email" Feature
### Describe the feature

- Add a "Resend Verification Link" button on the email verification page.
- Button should only be visible if the current verification link is expired.
- On click, it should trigger an API to send an email with new verification link to the user.
- Show a success or error toast based on the API response.

- Users can get stuck if their verification link expires.
- Improves overall user experience and reduces support requests.
- Ensures users can verify/complete account setup without friction.

### Confirmation

MukeshAbhi:
## Description

closes #766 

This PR introduces the **resend email verification** feature with rate limiting — allowing a maximum of **3 resend email requests per day**.

The feature integrates seamlessly into the **signin** and **verifyEmail** flows, improving user experience and ensuring smoother account activation handling.  

**No buttons are added to trigger resend verification.**

---

## What type of PR is this? 

- [x] 🍕 Feature   
- [x] ✅ Test  
 
---

### 🔹 Backend

- Added  **updateUserEmailCount** method in **UserService**  : Updates user resend email count.
- Added **updateUserToken** method in **UserTokenRepository** to handle token refresh.
- Added **resendVerificationEmail** method in **UserTokenService** to handle verification email resends with rate limiting and token refresh.
- Integrated this method into **signinController** and **verifyEmailController** for smoother user verification flow.
- Added tests for these.

### 🔹 Tests
- Updated existing tests and added new tests.
---

### Video 

https://github.com/user-attachments/assets/01ad466e-732e-4e92-9990-bd3e3855311f

 User services Tests 

image

User Token service Tests

image

 Signin Tests 
 

image

Verify Tests

image

### Coverage report 

image

image

---

## Checklist

### DIFF
--- a/packages/app/controllers/auth.js
+++ b/packages/app/controllers/auth.js
@@ -119,6 +119,7 @@ async function signupController(req, res, next) {
 async function signinController(req, res, next) {
   try {
     const userService = new UserService();
+    const userTokenService = new UserTokenService();
     let user = {};
     if (req.query.type === "guest") {
       user = await userService.getGuestUser();
@@ -140,12 +141,6 @@ async function signinController(req, res, next) {
           statusCode: 404,
         });
 
-      if (!user.is_verified)
-        return res.status(403).json({
-          error: STATUS_CODES[403],
-          message: Messages.EMAIL_NOT_VERIFIED,
-          statusCode: 403,
-        });
       if (user.is_deleted) {
         return res.status(400).json({
           error: STATUS_CODES[400],
@@ -160,6 +155,32 @@ async function signinController(req, res, next) {
           message: Messages.INCORRECT_EMAIL_PASS,
           statusCode: 404,
         });
+      }
+
+      if (!user.is_verified) {
+        try {
+          const result = await userTokenService.resendVerificationEmail(
+            user,
+            userService
+          );
+          return res.status(201).json({
+            message: result.message,
+            statusCode: 201,
+            source: "resendEmail",
+          });
+        } catch (err) {
+          let statusCode = 500;
+          let message = Messages.RESEND_EMAIL_FAILED;
+          if (err.statusCode) {
+            statusCode = err.statusCode;
+            message = err.message;
+          }
+          return res.status(statusCode).json({
+            message: message,
+            statusCode: statusCode,
+            source: "resendEmail",
+          });
+        }
       }
     }
     const currentDate = new Date();
@@ -256,14 +277,6 @@ async function verifyEmailController(req, res, next) {
       });
     }
 
-    if (userToken.isExpired()) {
-      return res.status(403).json({
-        error: STATUS_CODES[403],
-        message: Messages.EXPIRED_TOKEN,
-        statusCode: 403,
-      });
-    }
-
     const user = await userService.getUser(userToken.user_id);
     if (!user) {
       return res.status(404).json({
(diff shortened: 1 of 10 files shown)


packages/app/services/users.js
```diff
@@ -190,6 +190,25 @@ class UserService {
...
+  /**
+   * Updates the user's resend email count.
+   * @param {Object} user - The user.
+   * @param {Boolean} reset - Boolean value to reset the count.
+   * @returns - Returns updated user.
+   */
+  async updateUserEmailCount(user, reset = false) {
+    const updatedFields = {
+      last_verification_email_sent_at: new Date(),
+      resend_email_count: reset ? 1 : (user.resend_email_count || 0) + 1,
```
Hariom01010, line 203: in what case do we reset the user's resend email count? Also if we are resetting it is there any specfic reason for setting it to 1 and not 0 ?

MukeshAbhi: We will reset the email count once 24 hours have passed since the last email was sent. When the reset happens, the count will start at 1 (not 0), because an email is sent at the time of reset. This helps us keep an accurate count of the emails sent in the current 24-hour window.

packages/app/controllers/auth.js
```diff
@@ -161,6 +156,32 @@ async function signinController(req, res, next) {
...
+            statusCode: 201,
+            source: "resendEmail",
+          });
+        } catch (err) {
+          let statusCode = 500;
+          let message = Messages.RESEND_EMAIL_FAILED;
+          if (err.statusCode) {
+            statusCode = err.statusCode;
+            message = err.message;
+          }
```
Hariom01010, line 177: are we sending an error which doesn't have the statusCode field because of which we are setting default values for the statusCode and message?

MukeshAbhi: We are not sending any error without a statusCode field.
This ensures that if any unknown error occurs, it will automatically be assigned
statusCode = 500 and message = “Internal Server Error”. And also to distinguish between multiple errors we send.

packages/app/utils/constants.js
```diff
@@ -91,6 +91,10 @@ const Messages = {
   LOGO_REQUEST_CREATED: "Logo request created successfully",
   USER_ALREADY_HAS_PENDING: "You already have a pending request.",
   COMPANY_URL_ALREADY_PENDING: "This company url is already under review.",
+  TRY_AGAIN: "Try again after 24 hours.",
```
Hariom01010, line 94: nit: can we change the key name to something more meaningful

MukeshAbhi: Yes I will change. Please suggest me a name

packages/app/services/usertoken.js
```diff
@@ -41,6 +52,59 @@ class UserTokenService {
...
+      await userService.updateUserEmailCount(user, true);
+    } else if (user.resend_email_count >= 3) {
+      const error = new Error(Messages.TRY_AGAIN);
+      error.statusCode = 429;
+      throw error;
+    } else {
+      await userService.updateUserEmailCount(user, false);
+    }
+
+    const updatedToken = await this.updateUserToken(userToken.token);
```
Sumitgitup, line 86: The current workflow updates the token and counters before email delivery is confirmed. To prevent **inconsistent state** upon email failure, **only commit the changes (token and counters) after a successful send**.

MukeshAbhi: Thanks for pointing it out!
In our implementation, sendEmail always resolves (it logs errors internally and does not throw). Because of that, the main point where inconsistent state could occur is during updateUserToken, which must succeed before we send an email containing the new token URL.

To avoid updating counters in cases where the token update fails, I moved updateUserEmailCount to the end. This ensures that the user’s resend counters are only updated after the token update and the email-send step both complete.

If we later change sendEmail to throw errors, we can extend the consistency guarantees further.

Hariom01010 (approved):

PR #878, opened 2025-11-15
Feat/download button

### LINKED ISSUE #870: `0.5.0` Enable download button to get the user data
### Describe the feature

Enable the "Download " button in the user settings screen (SettingCard.jsx). This will allow an authenticated user to download a complete JSON file of all their personal data.

Image

This is a critical feature for data portability and user control, allowing users to have a machine-readable copy of their data.

### Confirmation

Sumitgitup:
## Description

Closes #870 

This PR enables the download data to let the user to download their data.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [x] 👨‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🛠️ CI/CD

https://github.com/user-attachments/assets/9046b849-e795-43ae-9f8b-f0e8764c4369

### Test Coverage

screenshot-2025-11-15_19-27-47

### Test cases

screenshot-2025-11-15_19-22-56

### API Response

screenshot-2025-11-15_19-20-54

## Checklist

### DIFF
--- a/packages/app/__tests__/controllers/auth/reset-password.js
+++ b/packages/app/__tests__/controllers/auth/reset-password.js
@@ -9,7 +9,6 @@ const mongoose = require("mongoose");
 const app = require("../../../server");
 const dummyPassword =
   require("../../../utils/generatePassword").generatePassword();
-jest.mock("jsonwebtoken");
 jwt.verify = jest.fn(() => ({ userId: new mongoose.Types.ObjectId() }));
 
 describe("RESET PASSWORD API", () => {
@@ -50,7 +49,10 @@ describe("RESET PASSWORD API", () => {
     const response = await request(app)
       .patch(ENDPOINTS.RESET_PASSWORD)
       .set("Cookie", ["resetPasswordSession=mockToken"])
-      .send({ newPassword: dummyPassword.repeat(3) });
+      .send({
+        newPassword: dummyPassword.repeat(3),
+        confirmPassword: "dummyPassword".repeat(3),
+      });
 
     expect(response.status).toBe(422);
     expect(response.body).toEqual({
--- a/packages/app/__tests__/controllers/users/downloaduserdata.js
+++ b/packages/app/__tests__/controllers/users/downloaduserdata.js
@@ -25,7 +25,7 @@ describe("Controller: downloadUserData", () => {
     UserService.prototype.getUserDataForDownload = jest.fn();
   });
 
-  it("should download user data on 200 happy path", async () => {
+  it("should successfully return user data as a JSON attachment", async () => {
     const mockData = { profile: { name: "Test User" } };
     UserService.prototype.getUserDataForDownload.mockResolvedValue(mockData);
 
--- a/packages/app/__tests__/controllers/users/downloaduserdata.js
+++ b/packages/app/__tests__/controllers/users/downloaduserdata.js
@@ -1,0 +1,82 @@
+const { downloadUserData } = require("../../../controllers/users");
+const UserService = require("../../../services/users");
+
+jest.mock("../../../services/users");
+
+describe("Controller: downloadUserData", () => {
+  let req, res, next;
+  const mockUserId = "user_123";
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+
+    req = {
+      userData: { userId: mockUserId },
+    };
+
+    res = {
+      status: jest.fn(() => res),
+      send: jest.fn(() => res),
+      setHeader: jest.fn(() => res),
+    };
+
+    next = jest.fn();
+
+    UserService.prototype.getUserDataForDownload = jest.fn();
+  });
+
+  it("should download user data on 200 happy path", async () => {
+    const mockData = { profile: { name: "Test User" } };
+    UserService.prototype.getUserDataForDownload.mockResolvedValue(mockData);
+
+    await downloadUserData(req, res, next);
+
+    expect(UserService.prototype.getUserDataForDownload).toHaveBeenCalledWith(
+      mockUserId
+    );
+
+    expect(res.status).toHaveBeenCalledWith(200);
+    expect(res.setHeader).toHaveBeenCalledWith(
+      "Content-Type",
+      "application/json"
+    );
+    expect(res.setHeader).toHaveBeenCalledWith(
+      "Content-Disposition",
+      `attachment; filename="user_data_${mockUserId}.json"`
+    );
+    expect(res.send).toHaveBeenCalledWith(JSON.stringify(mockData, null, 2));
+
+    expect(next).not.toHaveBeenCalled();
+  });
+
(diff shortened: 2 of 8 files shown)


packages/app/__tests__/controllers/auth/reset-password.js
```diff
@@ -9,7 +9,6 @@ const mongoose = require("mongoose");
 const app = require("../../../server");
 const dummyPassword =
   require("../../../utils/generatePassword").generatePassword();
-jest.mock("jsonwebtoken");
```
Hariom01010, line 12: why did we remove this?

packages/app/__tests__/controllers/users/downloaduserdata.js
```diff
@@ -0,0 +1,82 @@
...
+      send: jest.fn(() => res),
+      setHeader: jest.fn(() => res),
+    };
+
+    next = jest.fn();
+
+    UserService.prototype.getUserDataForDownload = jest.fn();
+  });
+
+  it("should download user data on 200 happy path", async () => {
```
Hariom01010: nit: the test case naming can be improved

packages/app/repositories/base.js
```diff
@@ -9,6 +9,10 @@ class BaseRepository {
     this.model = model;
   }
 
+  async find(query) {
+    return await this.model.find(query);
+  }
+
```
Hariom01010, line 15: any specific reason to introduce this method in base repository. Also i saw you are using it for the requests repository. So instead of defining a method here can't we introduce a function specifically in the requests repository?

packages/app/__tests__/controllers/auth/reset-password.js
Sumitgitup: We removed `jest.mock("jsonwebtoken")` because our new unit test calls the `downloadUserData `controller function directly, not through an HTTP request.

This new test bypasses the `authMiddleware`, so `JWT.verify` is never called. We just pass the user data directly in the mock `req `object instead.

packages/app/repositories/base.js
Sumitgitup: This change follows the **DRY** (Don't Repeat Yourself) principle.

The `BaseRepository` acts as a parent class that our other repositories, like `RequestRepository` and `UsersRepository`, inherit from.

By placing common methods like `find`, `getById`, and `create` in the `BaseRepository`, all child repositories get this functionality automatically. This avoids duplicating logic and keeps our codebase cleaner and easier to maintain

packages/app/__tests__/controllers/users/downloaduserdata.js
Sumitgitup: Thanks for pointing that out.
Any suggestion from your side ?

packages/app/__tests__/controllers/users/downloaduserdata.js
```diff
@@ -0,0 +1,82 @@
...
+    const mockData = { profile: { name: "Test User" } };
+    UserService.prototype.getUserDataForDownload.mockResolvedValue(mockData);
+
+    await downloadUserData(req, res, next);
```
MukeshAbhi, line 32: Most of the tests in the codebase use the HTTP request–response pattern, and since we also have a "/download-data" endpoint here, it would be better to follow the same approach for consistency. This keeps the tests aligned with the rest of the suite and ensures we are validating the actual API behavior rather than just internal function calls

packages/app/controllers/users.js
```diff
@@ -324,6 +324,33 @@ async function logoRequestController(req, res, next) {
   }
 }
 
+/**
+ * Controller function to download user data as JSON
```
MukeshAbhi: It would be helpful to make this comment a bit more descriptive so future readers can easily understand the intent behind this code

packages/app/controllers/users.js
```diff
@@ -324,6 +324,33 @@ async function logoRequestController(req, res, next) {
...
+async function downloadUserData(req, res, next) {
+  try {
+    const { userId } = req.userData;
+
+    const userService = new UserService();
+
+    const dataToDownload = await userService.getUserDataForDownload(userId);
+
+    if (!dataToDownload) {
+      return res.status(404).send({ message: "User not found." });
```
MukeshAbhi: Please use the existing messages from constants.js, or update them if needed.

MukeshAbhi (changes requested):

packages/app/routes/users.js
```diff
@@ -17,5 +18,6 @@ router.post("/api-key", authMiddleware(), generateKeyController);
 router.delete("/api-key/:keyId", authMiddleware(), destroyKeyController);
 router.put("/password", authMiddleware(), updatePasswordController);
 router.post("/request", authMiddleware(), logoRequestController);
+router.get("/download-data", authMiddleware(), downloadUserData);
```
aps08: /download

packages/app/__tests__/controllers/users/downloaduserdata.js
Sumitgitup: That's a great point, and I agree that consistency is important.

When I started, I found that all of our controller integration tests (like `createapikey.js`, `getuserdata.js`, and this one) are currently failing with a 500 error. This indicates a deeper problem with our `supertest` environment that's causing the server to crash.

I've already refactored the "fat" controller logic for `downloadUserData` into the `UserService`, which is a good cleanup regardless

packages/app/controllers/users.js
Sumitgitup: Thanks for pointing that out.

I'll use the existing messages from the `constants.js`.

packages/app/controllers/users.js
Sumitgitup: Yeah sure, I'll do that

packages/app/routes/users.js
Sumitgitup: Thanks for the suggestion

MukeshAbhi (approved):

packages/app/controllers/users.js
```diff
@@ -324,6 +324,35 @@ async function logoRequestController(req, res, next) {
...
+    const userService = new UserService();
+
+    const dataToDownload = await userService.getUserDataForDownload(userId);
+
+    if (!dataToDownload) {
+      return res.status(404).send({ message: Messages.USER_NOT_FOUND });
+    }
+
+    const fileName = `user_data_${userId}.json`;
+    const jsonContent = JSON.stringify(dataToDownload, null, 2);
```
MukeshAbhi, line 345: Using the username instead of userId for the file name may improve readability and make the file more identifiable. This is just my personal suggestion. ex: **openlogo_data_username**

MukeshAbhi (approved):

Hariom01010 (approved):

PR #884, opened 2025-11-26
Cache invalidation for cloudfront CDN

### LINKED ISSUE #860: `0.5.0` add: cache invalidation for cloudfront CDN
### Describe the feature

Add automated cache invalidation for the CloudFront CDN whenever content updates occur. This ensures that users always receive the latest assets and data from the CDN instead of outdated cached versions.

Stale CDN caches can lead to improper functionality, content mismatches for users. Automated invalidation ensures consistent content freshness, improves user experience, and reduces manual intervention after releases.

### Confirmation

MukeshAbhi:
## Description
Closes #860  

This PR ensures CloudFront cache is invalidated when an admin updates an existing image.

### Changes Included:
- Reused the existing `cloudFrontInvalidate` function to perform cache invalidation.
- Cache invalidation now occurs automatically when an admin updates an existing image.
- Updated the IAM policy to ensure the created IAM user has the required permissions to perform cache invalidation.

- [x] 🐛 Bug Fix  
- [x] 🔥 Performance Improvements  
- [x] ✅ Test  

https://github.com/user-attachments/assets/cad1f015-7b15-435f-91bf-50022b562c37

https://github.com/user-attachments/assets/7f447a16-a99a-4d10-a860-fc21cd775c0c

`updateMetaData.js`

image

## Checklist

### DIFF
--- a/packages/app/repositories/images.js
+++ b/packages/app/repositories/images.js
@@ -1,7 +1,10 @@
 const BaseRepository = require("./base");
 const Image = require("../models/images");
 
-const { cloudFrontSignedURL } = require("../utils/cloudFront");
+const {
+  cloudFrontSignedURL,
+  cloudFrontInvalidate,
+} = require("../utils/cloudFront");
 
 /**
  * The ImageRepository extends BaseRepository to manage ContactUs model operations, inheriting CRUD methods like getById, getAll, create, update, and delete..
@@ -72,6 +75,17 @@ class ImagesRepository extends BaseRepository {
   }
   async getImagesCount() {
     return await Image.countDocuments();
+  }
+
+  /**
+   * Invalidates the CloudFront cache for the specified image path.
+   *
+   * @param {string} imageUrl - The relative path of the image to invalidate.
+   * @returns {Promise<void>} Resolves when the cache invalidation request is completed.
+   */
+  async invalidateCloudFrontCache(imageUrl) {
+    const paths = [`/${imageUrl}`];
+    await cloudFrontInvalidate(paths);
   }
 }
 
--- a/packages/app/utils/cloudFront.js
+++ b/packages/app/utils/cloudFront.js
@@ -3,6 +3,7 @@ const {
   CloudFrontClient,
   CreateInvalidationCommand,
 } = require("@aws-sdk/client-cloudfront"); // CommonJS import
+const { CLOUD_FRONT_REGION } = require("./constants");
 
 const getSignedUrl = cloudfrontSigner.getSignedUrl;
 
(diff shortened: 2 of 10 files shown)


packages/app/utils/constants.js
```diff
@@ -99,6 +99,8 @@ const Messages = {
 
 const ExtractCompanyNameFromUrlRegex = /:\/\/(?:www\.)?([^./]+)\./i;
 
+const CLOUD_FRONT_REGION = "us-east-1";
```
Hariom01010, line 102: any specific region to hardcode this CLOUD_FRONT_REGION value to us-east-1?

packages/app/utils/cloudFront.js
```diff
@@ -39,16 +40,16 @@ function cloudFrontSignedURL(path) {
...
-  credentials: {
-    accessKeyId: process.env.ACCESS_KEY,
-    secretAccessKey: process.env.SECRET_ACCESS_KEY,
-  },
-};
-const client = new CloudFrontClient(config);
-
 async function cloudFrontInvalidate(paths) {
+  const config = {
+    region: CLOUD_FRONT_REGION,
```
Hariom01010, line 45: why did we replaced the BUCKET_REGION  with the CLOUD_FRONT_REGION? Also earlier we were storing that value in env variables but now we are storing the new variable in constant file. Will there be an issue if we use enviornment variables for it too?

packages/app/utils/constants.js
MukeshAbhi: CLOUD FRONT only works us-east-1

packages/app/utils/cloudFront.js
MukeshAbhi: Because cloud fronts only works on us-east-1. Yes if you use your bucket region it won't work.

packages/app/repositories/images.js
```diff
@@ -73,6 +76,17 @@ class ImagesRepository extends BaseRepository {
...
+  /**
+   * Invalidates the CloudFront cache for the specified image path.
+   *
+   * @param {string} imageUrl - The relative path of the image to invalidate.
+   * @returns {Promise<void>} Resolves when the cache invalidation request is completed.
+   */
+  async invalidateCloudFrontCache(imageUrl) {
+    const paths = [`/${imageUrl}`];
+    await cloudFrontInvalidate(paths);
+  }
```
Hariom01010, line 89: I saw we are not returning anything after invalidating. Does cloudfront confirms that invalidation never fails? If not what will happen in case invalidation fails?

MukeshAbhi: cloud front returns few fields like invalidation id etc but we don't use these information and amazon guarantees the process (99.99%) so not returning / accessing returned data

aps08 (approved):

PR #938, opened 2026-01-12
feat: JWT to session based auth

### LINKED ISSUE #916: `0.7.0` Auth migration from JWT to session-based
Currently we are using JWT based authentication based system from which we would be now transition or migrate to session based authentication.

Smayur0:
## Description

closes #916 
JWT to session based auth

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🛠️ CI/CD

## Checklist

### DIFF
--- a/packages/app/controllers/auth.js
+++ b/packages/app/controllers/auth.js
@@ -1,11 +1,12 @@
 const { STATUS_CODES } = require("http");
-const jwt = require("jsonwebtoken");
+
 const {
   UserService,
   SubscriptionService,
   UserTokenService,
   SendEmailService,
   UserSessionService,
+  PasswordResetService,
 } = require("../services");
 const {
   signupPayloadSchema,
@@ -112,7 +113,9 @@ async function signupController(req, res, next) {
 /**
  * This controller validates the sign-in payload, checks if the email exists,
  * verifies the user's email status, and compares the provided password with
- * the stored one. Set sets a JWT cookie and returns a successful response.
+ * the stored one.
+ * On successful authentication, it creates or reuses an active user session
+ * and sets a sessionId cookie to maintain a session-based login.
  *
  * @param {import("express").Request} req
  * @param {import("express").Response} res
@@ -183,25 +186,11 @@ async function signinController(req, res, next) {
       }
     }
 
-    // const existingDeviceId = req.cookies.deviceId;
-    // let device = {};
-    // device = await deviceInfoService.findDeviceById(existingDeviceId);
-
-    // if (!device) {
-    //   device = await deviceInfoService.createDevice({
-    //     userId: user._id,
-    //     deviceId: existingDeviceId,
-    //     userAgent: req.headers["user-agent"],
-    //     ipAddress: req.ip,
-    //   });
-    // }
-
     let session = {};
     session = await userSessionService.userActiveSession(user._id);
     if (!session) {
       session = await userSessionService.createSession({
         userId: user._id,
-        // deviceId: device.deviceId,
       });
     }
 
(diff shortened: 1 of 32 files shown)


packages/app/controllers/auth.js
```diff
@@ -223,15 +208,7 @@ async function signinController(req, res, next) {
       domain: getIsProduction() ? ".openlogo.fyi" : "localhost",
```
printgourav: Please take care of stage as well.

anujarora0502 (changes requested):
I have gone through the changes, I have few suggestions, this PR was marked for review when tests were failing, we should have checked locally before marking it for review, Sonarqube has found 3 new issues, let's fix them. Everything else looks good to me, let me know once my concerns are resolved.

packages/app/models/passwordResetSession.js
```diff
@@ -0,0 +1,47 @@
...
+    type: String,
+    required: true,
+  },
+
+  isActive: {
+    type: Boolean,
+    default: true,
+  },
+
+  createdAt: {
```
sachinkmrsin, line 32: why createdAt and not use mongodb timestamp ?

packages/app/models/userSession.js
```diff
@@ -0,0 +1,39 @@
...
+    type: Date,
+    default: Date.now,
+  },
+
+  expiresAt: {
+    type: Date,
+    required: true,
+    expires: 0, // TTL
+  },
+});
```
sachinkmrsin, line 37: can also include timestamp here .

packages/app/utils/constants.js
```diff
@@ -113,6 +113,11 @@ const CLOUD_FRONT_REGION = "us-east-1";
 const getIsProduction = () =>
   process.env.NODE_ENV?.trim().toLowerCase() === "prod";
 
+const USER_SAFE_FIELDS =
+  "name email role userId is_verified subscription_id created_at is_deleted updated_at ";
```
sachinkmrsin, line 117: why is userId here, the _id would be the userId right.

packages/ui/src/components/dropdown/UserDropDown.jsx
sachinkmrsin, line 1: frontend file added to backend pr.

sachinkmrsin (commented):
logic is good. just address the added comments.

packages/app/models/passwordResetSession.js
Smayur0: we are using mongoDB TTL to remove expired sessions for that we need expiredAt mongoDB not use timestamp for that.

packages/app/models/userSession.js
Smayur0: we are using mongoDB TTL to remove expired sessions for that we need expiredAt mongoDB not use timestamp for that.

packages/ui/src/components/dropdown/UserDropDown.jsx
Smayur0: I only changed one file, so I added it here. Since it’s my first PR to OpenLogo, should I open a separate PR for the Frontend change?

packages/app/utils/constants.js
Smayur0: yes, will update this

packages/app/models/passwordResetSession.js
sachinkmrsin: you dont get me , you expiredAt is fine.  mongo db models come with a configuration where if you set `timestamps : true` it will auto generate createdAt and updatedAt and maintain them. why create a createdAt and not use the feature.

packages/app/models/userSession.js
sachinkmrsin: even if you are going to delete the sessions,  having information on when sessions were created and updated will be helpful for the user right.

packages/ui/src/components/dropdown/UserDropDown.jsx
sachinkmrsin: i believed this was a pure backend pr so i addressed it .

sachinkmrsin:
Also address the minor sornarQube errors.

Issue #943, opened 2026-01-26
`0.8.0` Logo Usage Rewards


BansalAbhinav:
### Describe the feature

- Add a usage-based rewards system for OpenLogo where logo creators earn rewards only when their logos are used by Pro plan users via API.
- Currently, uploads are not rewarded based on real usage, which leads to low-quality uploads and unfair incentives.

### It should be implemented because ? Additional context

- Rewards creators based on actual paid usage, not uploads
- Encourages higher-quality logos
- Prevents abuse (self-usage, spam requests, duplicate counts)
- Creates a reliable base for future monetization
- Only Pro users count toward rewards
- Each user is counted once per logo
- Hobby usage is logged but not rewarded
- Rewards are calculated asynchronously and fully auditable

### Confirmation

- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand this documentation issue.

PR #963, opened 2026-02-23
Feat/reward system backend

sachinkmrsin:
## Description
closes:#943

This is a test only pr created to distribute the size of the pr. Contains test for reward creation and transactions. 
Tech Document : https://www.notion.so/OpenLogo-Reward-System-Architecture-2e11cb09a77180ed8e60d4f57e6d0d49?source=copy_link

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🛠️ CI/CD

## Checklist


packages/app/__tests__/controllers/rewards.js
```diff
@@ -0,0 +1,862 @@
...
+          userId: MOCK_USERS[0]._id.toString(),
+          points: 50,
+        });
+
+      expect(response.status).toBe(401);
+    });
+
+    it("400 - Missing required fields", async () => {
+      const adminSession = {
+        userId: MOCK_USERS[2], // Admin user
```
AryaDharkar: userID has entire user object passed to it?

packages/app/__tests__/controllers/rewards.js
```diff
@@ -0,0 +1,862 @@
...
+      };
+
+      jest
+        .spyOn(UserSessionService.prototype, "validateSession")
```
AryaDharkar: Admin sessions are built as { userId: MOCK_USERS[2] } (full object), but non-admin as { userId: MOCK_USERS[0] } (also full object). If the controller does session.userId._id vs session.userId, one path will throw. The shape should be consistent and match what validateSession actually returns in production.

packages/app/__tests__/controllers/rewards.js
```diff
@@ -0,0 +1,862 @@
...
+        .mockResolvedValue(adminSession);
+
+      const response = await request(app)
+        .post(endpoint)
+        .set("Cookie", `sessionId=${MOCK_SESSION_ID}`)
+        .send({
+          imageId: MOCK_IMAGES[0]._id.toString(),
+          userId: MOCK_USERS[0]._id.toString(),
+          points: -5, // Negative points
```
AryaDharkar: If we have to check >0, please also check if 0 is inclusive in addition to -5.

packages/app/__tests__/controllers/rewards.js
```diff
@@ -0,0 +1,862 @@
...
+
+      const response = await request(app)
+        .get(endpoint)
+        .set("Cookie", `sessionId=${MOCK_SESSION_ID}`);
+
+      expect(response.status).toBe(500);
+    });
+  });
+
+  describe("getRewardsLeaderboardController", () => {
```
AryaDharkar: No test checks what happens when limit is 0, negative, or non-numeric. The controller has to reject these values then

packages/app/__tests__/services/rewards.test.js
AryaDharkar, line 1: Please test the service logic itself. RewardsService is the thing under test, not a dependency. When you mock RewardsService.prototype.processRewardsForImage and then call rewardsService.processRewardsForImage(), you've replaced the real function with a mock one and are calling the mock function. Actual service code is not running.

packages/app/__tests__/services/rewards.test.js
```diff
@@ -0,0 +1,246 @@
...
+      const imageId = new mongoose.Types.ObjectId();
+
+      jest
+        .spyOn(RewardsService.prototype, "processRewardsForImage")
+        .mockResolvedValue(MOCK_PROCESS_REWARDS_SINGLE_MILESTONE);
+
+      const result = await rewardsService.processRewardsForImage(imageId);
+
+      expect(result.success).toBe(true);
+      expect(result.newMilestones).toBeDefined();
```
AryaDharkar: only checks result.newMilestones is defined, not its length or content

packages/app/__tests__/services/rewardTransactions.test.js
```diff
@@ -0,0 +1,765 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks();
+
+    rewardTrackingService = new RewardTrackingService();
+
+    mockLogoRequestLogsRepository = LogoRequestLogsRepository.mock.instances[0];
+    SubscriptionsRepository.mock.instances[0];
+    UsersRepository.mock.instances[0];
+
+    if (mockLogoRequestLogsRepository) {
```
AryaDharkar: These checks will silently be skipped if LogoRequestLogsRepository.mock.instances[0] is ever undefined or fails to launch an instance through beforeEach();

AryaDharkar (changes requested):
There is one fundamental error in the services tests. Please check

packages/app/__tests__/controllers/rewards.js
sachinkmrsin: it took some digging to find out , yes we have a function in the new sesssion based auth where the userId is populated by the user object thus it is so.

packages/app/__tests__/controllers/rewards.js
sachinkmrsin: the newly created auth module requires it to be so

packages/app/__tests__/controllers/rewards.js
sachinkmrsin: Yes 0 is included

packages/app/__tests__/controllers/milestones/delete.test.js
```diff
@@ -0,0 +1,123 @@
+/**
+ * MilestoneConfig Controller – Delete tests
+ * DELETE /api/admin/milestones/:id
+ */
+
+const request = require("supertest");
+const { STATUS_CODES } = require("http");
```
AryaDharkar: fix this. node:http is the latest convetion

packages/app/__tests__/services/rewardTransactions.test.js
```diff
@@ -0,0 +1,738 @@
...
+  let mockLogoRequestLogsRepository;
+  let mockImagesRepository;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+
+    rewardTrackingService = new RewardTrackingService();
+
+    mockLogoRequestLogsRepository = LogoRequestLogsRepository.mock.instances[0];
+    SubscriptionsRepository.mock.instances[0];
```
AryaDharkar: assign it to a var or remove if not getting used.

packages/app/__tests__/services/rewardTransactions.test.js
```diff
@@ -0,0 +1,738 @@
...
+        },
+        {
+          type: SubscriptionTypes.PRO,
+          shouldBeEligible: true,
+          reason: "VALID",
+        },
+      ];
+
+      for (const testCase of testCases) {
+        jest.clearAllMocks();
```
AryaDharkar: the test is resetting mocks within the loop, which can cause inconsistent behavior. Put it in beforeEach

packages/app/__tests__/services/rewards.test.js
```diff
@@ -0,0 +1,289 @@
...
+} = require("../../utils/mocks");
+
+describe("Reward System", () => {
+  let rewardTrackingService;
+  let rewardsService;
+
+  beforeEach(() => {
+    rewardTrackingService = new RewardTrackingService();
+    rewardsService = new RewardsService();
+    jest.clearAllMocks();
```
AryaDharkar, line 36: The services are initialized but their repositories aren't mocked. The tests spy on repository methods directly, but there's no jest.mock() at the top level like in rewardTransactions.test.js. Mock the repositories

packages/app/__tests__/controllers/users/getuserdata.js
```diff
@@ -133,8 +135,6 @@ describe("GETUSERDATA", () => {
       keys: [
         {
           _id: mockKeyModel._id.toString(),
-          key_description: mockKeyModel.key_description,
```
AryaDharkar, line 136: These are spied on above in the file. Remove the spy if not being used

packages/app/__tests__/controllers/logo/get-logo.js
```diff
@@ -1,16 +1,24 @@
...
 
 const {
   ImageService,
   KeyService,
   SubscriptionService,
   UserService,
+  RewardTransactionsService,
 } = require("../../../services");
 
+jest.mock("../../../services/rewardTransactions");
```
AryaDharkar, line 20: external dependencies for unit testing not added. If someone tests the file, its gonna throw errors other than integration testing. Add the dependency mocks just in case

packages/app/__tests__/services/rewards.test.js
```diff
@@ -0,0 +1,289 @@
...
+
+      const result = await rewardsService.reverseTransaction(
+        transactionId,
+        adminId,
+        "Abuse detected"
+      );
+
+      expect(result.is_reversed).toBe(true);
+      expect(result.points_awarded).toBe(
+        MOCK_UNREVERSED_TRANSACTION.points_awarded
```
AryaDharkar, line 309: object doesn't guarantee points_awarded exists.

packages/app/__tests__/services/rewards.test.js
```diff
@@ -0,0 +1,289 @@
...
+      // 5 new unique Pro users → milestone 5 (10 pts)
+      setupProcessRewardsMocks(
+        MOCK_ELIGIBLE_LOG_ENTRIES_5,
+        createMockRewardRecordEmpty()
+      );
+
+      const result =
+        await rewardsService.processRewardsForImage(REWARD_TEST_IMAGE_ID);
+
+      expect(result.success).toBe(true);
```
AryaDharkar, line 174: if result fails, test fails. Verify actual response structure, not just success

AryaDharkar (changes requested):
In addition to comments, make sonarqube changes too as they are following latest conventions

PR #977, opened 2026-03-09
fix: Enhance password reset and update logic to prevent same password…

### LINKED ISSUE #975: Prevent Users From Reusing Old Password During Password Reset
### Describe your issue

Currently, when a user resets their password, the system allows them to set the same password as their previous (old) password. This behavior weakens security because the password is not actually being changed.

### Steps to reproduce

1.Go to the Reset Password functionality in the application.

2.Enter a password in forgot password

3.In the New Password field, enter the same password that was previously used.

4.Submit the password reset request.

### Confirmation

rishang14:
## Description
 Closes #975 
 
 Summary

This PR introduces validation in the password reset flow to prevent users from setting a new password that is the same as their current password.

Problem

 during the password reset process, the system allowed users to reset their password to the same value as their existing password.

Solution

A password comparison check has been added before updating the user's password. The system now verifies whether the provided new password matches the current hashed password stored in the database.

If the new password matches the existing password:

The reset request is rejected.

A 400 response is returned with an appropriate error message.

Implementation Details

- [ ] 🍕 Feature
- [✔️ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [✔️ ] ✅ Test
- [ ] 🛠️ CI/CD

Screenshot 2026-03-10 212543

## Checklist
- [✔️ ] I have performed a self-review of my code  
- [ ✔️] I have added tests that prove my fix is effective or that my feature works  
- [ ✔️] New and existing unit tests pass locally with my changes

### DIFF
--- a/packages/app/__tests__/controllers/auth/reset-password.js
+++ b/packages/app/__tests__/controllers/auth/reset-password.js
@@ -1,6 +1,7 @@
 const request = require("supertest");
 const { STATUS_CODES } = require("http");
 const { ENDPOINTS } = require("../../../utils/testconstants");
+const { Users } = require("../../../models");
 const {
   UserService,
   UserTokenService,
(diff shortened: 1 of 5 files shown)


anujarora0502 (approved):
I have gone through the changes, It looks good to me! The sonarqube error is due to wrong jsdocs, we will need to fix them eventually to make sure sonarqube won't give us false results.

Issue #1011, opened 2026-04-23
Fix the analytics bug


Dhirenderchoudhary:
### Describe your issue

currently for a month we should be getting the data of current month, and same for the week, and that is not the case as of now, from backend as well we are getting correct data for week but not for month, month data should start from 1st of currrent month and end at the last of current month as it is there for the week. We need to confirm this if it is correct for the week or not as well.

## Parent Issue
#997 

### Confirmation

- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand this documentation issue.

Issue #997, opened 2026-04-14
[BACKLOG] List of backlog items


anujarora0502:
Pick your next task from these issues and make sure to remove it from the list once you have picked it up and created an issue for it -

INFO: Use google stitch for creating UI designs.

1. Revamp of home page as per this design pattern + navbar + footer - https://openlogo-design.vercel.app/
2. Update the get in touch modal as per the new design pattern - https://openlogo-design.vercel.app/
3. Revamp of operator dashboard, currently we are following the old design pattern there.
4. Update request logo modal as per the new design pattern - https://openlogo-design.vercel.app/
5. Revamp of https://www.openlogo.fyi/privacy page as per the new design pattern - https://openlogo-design.vercel.app/
6. Sign up form has password strength option, that should be red if strength is low, orange when medium, green on last two bars which means strong.
7. Revamp of release page - https://openlogo.fyi/release, as per the new design pattern - https://openlogo-design.vercel.app/, Currently this page requires code changes, but that should change, the admin should be able to add the release notes through dashboard along with the image.
8. SEO optimizations, refer this - https://github.com/TeamShiksha/tech-spec/blob/prod/RSVP/SEO.pdf

PR #1038, opened 2026-05-11
[Fix] Issue-1011 Correct analytics date ranges, and the PR description must include Fixes

Dhirenderchoudhary:
## Description

Currently for a month we should be getting the data of current month, and same for the week, and that is not the case as of now, from backend as well we are getting correct data for week but not for month, month data should start from 1st of currrent month and end at the last of current month as it is there for the week. We need to confirm this if it is correct for the week or not as well.

## Issue:
#1011
#997 

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🛠️ CI/CD

## Checklist

### DIFF
--- a/packages/app/__tests__/controllers/logoRequestLogs.js/getStats.js
+++ b/packages/app/__tests__/controllers/logoRequestLogs.js/getStats.js
@@ -30,7 +30,7 @@ describe("GET LOGO REQUEST STATS", () => {
   });
 
   describe("Weekly Stats", () => {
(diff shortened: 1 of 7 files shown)


anujarora0502 (approved):
Tests are passing, code looks good as well, I couldn't find any issues!

PR #1043, opened 2026-05-18
[Feat] Issue-1007 Sign up password strength ui

### LINKED ISSUE #1007: Sign up form has password strength
### Describe the enhancement

The current sign up form while making a password it doesn't shows password strength  

If user creates password it will show   red if strength is low, orange when medium, green on last two bars which means strong

### Confirmation

AditSingh7:
## Description

Closes #1007

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🛠️ CI/CD

## Checklist


sachinkmrsin:
link you issue number in your pr document as well like so "closes #number"

packages/ui/src/utils/Helpers.js
sachinkmrsin, line 1: why did we remove all the password rule checks ? are we not enforcing a strong password anymore?

sachinkmrsin (commented):
address comments, code practices followed. looks good to me.

packages/ui/src/utils/Helpers.js
AditSingh7: This only removes the password error messages shown below the password field. Password validation still works through the live requirement checklist, and only the “Password is required” message is kept.

packages/ui/src/utils/Helpers.js
```diff
@@ -18,46 +14,7 @@ export const isValidPassword = (password) => {
...
-          !PASSWORD_RULES.specialCharRegex.test(password),
-        message: PASSWORD_VALIDATION_MESSAGES.specialChar,
-      },
-    ];
-
-    checks.forEach(({ condition, message }) => {
-      if (condition) {
-        errors.password = message;
-      }
-    });
```
anujarora0502, line 60: What sachin said, it is absolutely correct, we are removing the validation from frontend, we are making unnecessary calls to backend and the error is coming from backend, we should reject the sign up call beforehand.

packages/ui/src/components/auth/Signup.jsx
```diff
@@ -49,7 +49,7 @@ function SignUp({ toggleForm, onClose }) {
       setFormErrors({
         [focusedField]: validationErrors[focusedField] || "",
       });
-    }, 500);
+    }, 2000);
```
anujarora0502: Why did we increase it?

anujarora0502 (changes requested):
Tests are failing we will need to fix it, please take a look. And also the sign up form is getting really big, let's change the design to this.

Image

packages/ui/src/components/auth/Signup.jsx
AditSingh7: When we click the password field the requirements animation is shown and then instantly another animation the red line error message shown below the password field , it looks a bit weird so i increased the time to 2 second and another reason i increased time because when user about to make a password at least it should wait for a bit .

anujarora0502 (commented):
Remove these messages, we are already showing it on the sign up form.

Image


Increase the shadow as currently the pointing popup is not visible.

Image

anujarora0502:
We need different shadows in dark and light mode, the issue that was coming in light mode is not in dark mode, so we will need to fix that. @AditSingh7

anujarora0502 (approved):
Looks good now.

PR #1046, opened 2026-05-21
[Feat] Issue-1004 subscription admin api

### LINKED ISSUE #1004: Allow Admins to view and update customer's subscrition status
### Describe the feature

Allows the admin users to view customer's data securely and update their subscription if required without any payment.

Required for testing of rewards features.

### Confirmation

sachinkmrsin:
## Description
closes: #1004

- Implemented changeSubscriptionPlanSchema for validating subscription changes.
- Added changeSubscriptionPlanController to handle subscription upgrades/downgrades.
- Created listUsersController for retrieving customer users with subscription details.
- Introduced SubscriptionLogRepository for logging subscription changes.
- Updated routes for admin user subscription management.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🛠️ CI/CD

image

## Checklist


packages/app/controllers/subscriptions.js
```diff
@@ -0,0 +1,134 @@
...
+        user.subscription_id,
+        plan
+      );
+
+    await subscriptionService.createSubscriptionLog({
+      user_id: user._id,
+      subscription_id: user.subscription_id,
+      from_plan: subscription.type,
+      to_plan: plan,
+      changed_by: req.userData.userId,
```
AryaDharkar: the model expects an ObjectId, not a string. Type mismatch: storing string where ObjectId is expected

packages/app/utils/mocks.js
AryaDharkar, line 1: constants.js only exports DefaultSubscriptionPlan. The ProSubscriptionPlan is defined but NOT exported.

packages/app/repositories/users.js
```diff
@@ -50,6 +50,83 @@ class UsersRepository extends BaseRepository {
...
+      role: "CUSTOMER",
+    };
+
+    if (!includeDeleted) {
+      matchStage.is_deleted = false;
+    }
+
+    if (search) {
+      const escapedSearch = search.replaceAll(
+        /[.*+?^${}()|[\]\\]/g,
```
AryaDharkar: It should be `[.*+?^${}()|[\\\]\\]`  to properly escape brackets for queries with brackets. Very rare edge case but better if fixed. Thx

packages/app/repositories/subscriptionLog.js
```diff
@@ -0,0 +1,33 @@
...
+        .find({})
+        .populate("user_id", "name email")
+        .populate("changed_by", "name email")
+        .sort({ createdAt: -1 })
+        .skip(skip)
+        .limit(limit)
+        .select("-__v"),
+      this.model.countDocuments({}),
+    ]);
+    return { logs, total, totalPages: Math.ceil(total / limit) };
```
AryaDharkar, line 29: Division by zero edge case check required for limit==0

packages/app/controllers/subscriptions.js
```diff
@@ -0,0 +1,134 @@
...
+      });
+    }
+
+    const updatedSubscription =
+      await subscriptionService.changeSubscriptionPlan(
+        user.subscription_id,
+        plan
+      );
+
+    await subscriptionService.createSubscriptionLog({
```
AryaDharkar: `req.userData.userId ` comes from the session, but there's no validation that the admin user actually still exists in the database. If admin is deleted but session persists, the audit log will have a reference to a non-existent user

packages/app/controllers/subscriptions.js
```diff
@@ -0,0 +1,134 @@
...
+    const user = await usersService.getUser(userId);
+    if (!user) {
+      return res.status(404).json({
+        statusCode: 404,
+        error: STATUS_CODES[404],
+        message: Messages.USER_NOT_FOUND,
+      });
+    }
+
+    const subscription = await subscriptionService.getSubscription(
```
AryaDharkar: Between reading the subscription (line 51) and checking if it's the same plan (line 62) and updating (line 71), another admin could update the same user's subscription. The audit log would have incorrect from_plan value.
fix would be use database-level compare-and-swap or fetch just before update

packages/app/controllers/subscriptions.js
```diff
@@ -0,0 +1,134 @@
...
+      user.subscription_id
+    );
+    if (!subscription) {
+      return res.status(404).json({
+        statusCode: 404,
+        error: STATUS_CODES[404],
+        message: "Subscription not found.",
```
AryaDharkar: All other error messages use `Messages `constants from `utils/constants.js`, but this one is hardcoded. Inconsistent pattern and harder to maintain.

packages/app/controllers/subscriptions.js
```diff
@@ -0,0 +1,134 @@
...
+      });
+    }
+
+    if (subscription.type === plan) {
+      return res.status(400).json({
+        statusCode: 400,
+        error: STATUS_CODES[400],
+        message: Messages.PLAN_ALREADY_ACTIVE,
```
AryaDharkar: The Messages object doesn't include:

```
PLAN_ALREADY_ACTIVE - used in subscriptions.js line 66
PLAN_CHANGE_SUCCESS - used in subscriptions.js line 87
```
These are referenced but not defined. They must be added to constants.js based on frontend PR file changes

packages/app/schemas/admin.js
```diff
@@ -20,4 +20,32 @@ const imageReuploadSchema = Joi.object().keys({
...
+
+const listUsersQuerySchema = Joi.object({
+  search: Joi.string().trim().max(100).optional(),
+  page: Joi.number().integer().min(1).default(1),
+  limit: Joi.number().integer().min(1).max(100).default(20),
+  includeDeleted: Joi.boolean().default(false),
+});
+
+const listSubscriptionLogsQuerySchema = Joi.object({
+  page: Joi.number().integer().min(1).default(1),
```
AryaDharkar, line 41: Defaults to limit=20 but frontend hardcodes limit=10. This means if frontend doesn't send limit param, backend returns 20 items instead of expected 10. Inconsistent defaults.

AryaDharkar (changes requested):
use node('http') as per latest conventions. Note inconsistencies in constants.js and ensure all added fields are exported

packages/app/controllers/subscriptions.js
sachinkmrsin: how can an admin login and change a user's plan and not exist as the same time?

packages/app/controllers/subscriptions.js
sachinkmrsin: we can use transactions for it.

packages/app/utils/mocks.js
sachinkmrsin: exported on line 157

packages/app/repositories/subscriptionLog.js
sachinkmrsin: we have JOI object to verify limit is not negative.

anujarora0502 (approved):
Not reviewing it again, as it has been reviewed already.

PR #1070, opened 2026-07-08
[Docs] Issue-943 reward system docs

sachinkmrsin:
## Description

Docs pr for the reward system and related api's. Contains the postman collection as well as well some cleanup code required for better error handling.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [x] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🛠️ CI/CD

## Checklist


packages/app/controllers/rewards.js
0-mstrmind, line 109: `userId` should be validated before querying result

0-mstrmind, line 1: some field which we are taking from user should be validated like `transactionId` (example at line no. 188 transaction ID is not validated similar in other controllers) provided by the user or not.

0-mstrmind, line 196: if there is no transactions previously I think it may occur an error what if it returns `null`. Should also be implemented in other controllers also

packages/app/controllers/milestoneConfig.js
0-mstrmind, line 122: validation check of `!config` should be there

0-mstrmind, line 183: `!config` check is missing

0-mstrmind, line 276: `!config` check is missing

0-mstrmind, line 335: `!config` check is missing

0-mstrmind, line 54: `req.params.id` should be validated, it can be `null` hence we are asking this from user. Also, should be fix in all over the file

PR #1088, opened 2026-08-03
[Fix] Issue-1084 Fix logo URL button copying functionality

### LINKED ISSUE #1084: Replace the logo display with a Copy Link button that copies the logo URL with the API key as a placeholder.
### Describe the enhancement

Currently when someone searches for a logo, we just show the logo, instead we should add a button that copies the link for that logo with api key as placeholder.

_No response_

### Confirmation

AditSingh7:
## Description

fixes #1084 

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📄 Documentation Update
- [ ] 👨‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🛠️ CI/CD

## Checklist
- [ ] I have performed a self-review of my code  

### DIFF
--- a/packages/ui/src/components/demo/Demo.jsx
+++ b/packages/ui/src/components/demo/Demo.jsx
@@ -5,7 +5,11 @@ import { BUTTON_TEXT, DEMO, TILTED_BRANDS } from "../../utils/Constants.js";
 import styles from "./Demo.module.css";
 import Button from "../common/button/Button.jsx";
 import PropTypes from "prop-types";
-import { firstLetterCapitalString } from "../../utils/Helpers.js";
+import {
+  firstLetterCapitalString,
+  getBaseApiUrl,
+  getLogoDomain,
+} from "../../utils/Helpers.js";
 import { useApi } from "../../hooks/useApi.js";
 import LogoRequestForm from "./LogoRequestForm.jsx";
 import { AuthContext } from "../../contexts/Contexts.jsx";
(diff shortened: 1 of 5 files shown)


packages/ui/src/utils/Helpers.js
```diff
@@ -323,6 +323,20 @@ export const getBaseApiUrl = (domain) => {
...
+ *
+ * @param {string} imageUrl - Logo asset URL, e.g. "https://.../png/google.com.png".
+ * @param {string} companyName - Company name without the TLD, used as a fallback.
+ * @returns {string} - The domain (e.g. "google.com") when the asset name holds
+ * one, otherwise the given company name.
+ */
+export const getLogoDomain = (imageUrl, companyName) => {
+  const fileName = (imageUrl || "").split("?")[0].split("/").pop() || "";
+  const domain = fileName.replace(/\.[^.]+$/, "");
+  return domain.includes(".") ? domain : companyName;
```
harshith-venkatesh: Could we think of simplifying it by adding the check in the earliest for imageUrl here, as if the imageUrl is empty we would be returning companyName correct?

packages/ui/src/components/demo/Demo.jsx
```diff
@@ -42,6 +47,23 @@ const Demo = ({ openAuthModal }) => {
...
+    []
+  );
+
+  const buildLogoUrl = ({ image, companyName }) =>
+    `${baseApiUrl}/logo?key=${getLogoDomain(image, companyName)}&API_KEY=YOUR_API_KEY`;
+
+  const handleCopyLink = (company) => {
+    navigator.clipboard.writeText(buildLogoUrl(company)).then(() => {
+      setCopiedCompany(company.companyName);
+      setTimeout(() => setCopiedCompany(null), 2000);
```
harshith-venkatesh, line 64: what is the purpose of setting the copiedCompany to null here

packages/ui/src/utils/Helpers.js
AditSingh7: will add an early check for imageUrl and yes if imageUrl is empty this will return companyName

packages/ui/src/components/demo/Demo.jsx
AditSingh7: it clears the "Copied!" after 2s so the button doesn't stay forever on it.

anujarora0502 (approved):
Looks good!

